PR #1, opened 2016-05-02
Inf 39

vguzmanp:
* Work around a bug when compiling taglib in Trusty
* Create a `base-crypto` package
* Move packages to cerbero from cerbero-extras


recipes/taglib.recipe
```diff
@@ -30,6 +30,11 @@ class Recipe(recipe.Recipe):
         if self.config.target_platform == Platform.ANDROID:
             # configure for android
             self.configure_options += ' -DANDROID_NDK=1 '
+        if self.config.target_platform == Platform.LINUX:
```
ylatuya: This should be done only for Trusty, other distros might install libGL in a different path. Also link to the bug in case someone needs to understand the hack

packages/base-crypto.package
```diff
@@ -0,0 +1,23 @@
...
+    vendor = 'GStreamer Project'
+    org = 'org.freedesktop.gstreamer'
+    uuid = 'b3a6046d-5a39-4d54-93ee-3770d2f316f5'
+    deps = ['base-system']
+
+    files = ['gmp:libs', 'gnutls:libs:libs_extra:lang',
+             'libtasn1:libs', 'nettle:libs', 'openssl:libs']
+
+    platform_files_devel = {
+        Platform.ANDROID: ['ca-certificates:etc']
```
ylatuya: I don't think we have a ca-certificates recipe.

packages/gstreamer-networking.package
```diff
@@ -53,8 +53,7 @@ class Package(package.Package):
                  DistroVersion.UBUNTU_RARING,
                  DistroVersion.FEDORA_16,
                  DistroVersion.FEDORA_17]:
-            self.files += ['gmp:libs', 'gnutls:libs:libs_extra:lang',
-                           'libtasn1:libs', 'nettle:libs',
+            self.files += ['gmp:libs',
```
ylatuya: gmp:libs is provided by base-crypto

ylatuya:
:+1:

PR #2, opened 2016-05-23
Inf 44

ylatuya:


recipes/libxml2.recipe
```diff
@@ -20,16 +20,16 @@ class Recipe(recipe.Recipe):
 
 
     def prepare(self):
-        self.configure_options = '--with-docbook=no'
+        self.configure_options = ' --with-docbook=no '
         if self.config.target_platform == Platform.WINDOWS:
-            self.configure_options += '--without-threads'
+            self.configure_options += ' --without-threads '
         elif self.config.target_platform == Platform.ANDROID:
```
turran: Wouldn't make sense to have every --with-python=no configure flag be based on the nopython variant?

PR #3, opened 2016-05-25
OPTC-5

dfernandezramos:


turran:
The commits can be squashed to not include the deps package
Also, why is an arm related commit on the enums.py?

dfernandezramos:
@turran I was getting errors on cerbero (something like "Architecture has no attribute 'ARM64'") so I defined that arm related commit on the enums.py

turran:
:+1:

packages/oneplay-transcoder-deps.package
```diff
@@ -0,0 +1,32 @@
...
+        #     if not prefix.endswith('/'):
+        #         prefix += '/'
+        #     files = []
+        #     for x in r:
+        #         p = os.path.realpath(os.path.join(prefix, x))
+        #         files.append(p.replace(prefix, ''))
+        #         config = p+'.config'
+        #         if os.path.exists(config):
+        #             files.append(config.replace(prefix, ''))
+        #     r = list(set(r+files))
```
ylatuya: remove commented code

ylatuya:
Since cerbero is the open source part of the repository, the package packages/oneplay-transcoder-deps.package should be moved to fluendo-cerbero

ylatuya:
👍

PR #8, opened 2016-07-06
Add patch for -as-needed flag in libtool

vguzmanp:
It was adding the as-needed flag at the end, when it needs to be before the libs to link
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=347650


turran:
:+1:

PR #22, opened 2016-10-30
Inf 68

turran:


recipes/libxslt.recipe
```diff
@@ -12,3 +12,8 @@ class Recipe(recipe.Recipe):
     files_devel = ['bin/xsltproc%(bext)s', 'bin/xslt-config', 'include/libexslt', 'include/libxslt',
                    'lib/xsltConf.sh',
                    'lib/pkgconfig/libexslt.pc', 'lib/pkgconfig/libxslt.pc']
+
+    def prepare(self):
+        if self.config.platform != self.config.target_platform:
+            self.configure_options += ' --with-libxml-include-prefix=%s/include/libxml2' % self.config.prefix
+            self.configure_options += ' --with-libxml-lib-prefix=%s/lib' % self.config.prefix
```
ylatuya: use self.config.libdir instead

cerbero/build/cookbook.py
```diff
@@ -177,7 +177,8 @@ def update_step_status(self, recipe_name, step):
         @type step: str
         '''
         status = self._recipe_status(recipe_name)
-        status.steps.append(step)
+        if not step in status.steps:
+            status.steps.append(step)
```
ylatuya: Just to know, does it fixes the spurious "step already done" when all steps where done?

cerbero/packages/packagesstore.py
```diff
@@ -166,6 +166,19 @@ def _load_packages(self):
...
+        for recipe in self.cookbook.get_recipes_list():
+            p = self._package_from_recipe(recipe)
+            self._packages[p.name] = p
+
+    def _package_from_recipe(self, recipe):
+        p = package.Package(self._config, self, self.cookbook)
+        p.name = recipe.name
+        p.license = recipe.licenses
+        p.version = '%s-%s' % (recipe.version, self.cookbook.recipe_file_hash(recipe.name))
+        p.files = ['%s:libs' % p.name]
```
ylatuya: I would do it differently, the recipes hace their own functions to list files, and we want packages to contain the same files as the recipe, otherwise extracting a binary would install an uncomplete list of files.
I would create a new RecipePackage that takes a recipe in the contstructor
and override list_files and list_devel_files with recipe.dist_files_list () and recipe.devel_files_list()

cerbero/commands/package.py
```diff
@@ -44,7 +44,7 @@ def __init__(self):
             ArgparseArgument('-f', '--force', action='store_true',
                 default=False, help=_('Delete any existing package file')),
             ArgparseArgument('-d', '--no-devel', action='store_false',
-                default=True, help=_('Do not create the development version '
+                default=False, help=_('Do not create the development version '
```
ylatuya: you also need to invert action to store_true I think

cerbero/build/fridge.py
```diff
@@ -0,0 +1,123 @@
+# cerbero - a multi-platform build system for Open Source software
+# Copyright (C) 2012 Andoni Morales Alastruey <ylatuya@gmail.com>
```
ylatuya: Fix CR

cerbero/build/fridge.py
```diff
@@ -0,0 +1,123 @@
...
+
+from cerbero.errors import PackageNotFoundError, BuildStepError, FatalError
+from cerbero.utils import _, shell
+from cerbero.build.recipe import Recipe, BuildSteps
+from cerbero.utils import messages as m
+from cerbero.utils.shell import upload_curl, download_curl
+from cerbero.packages.disttarball import DistTarball
+from cerbero.packages import PackageType
+
+class Fridge (object):
```
ylatuya: This Fridge should be used as a Mixin to extend a Recipe, like Source or Build blocks extend a recipe

cerbero/build/oven.py
```diff
@@ -36,57 +36,47 @@ class Oven (object):
...
-            recipes = [recipes]
-        self.recipes = recipes
         self.cookbook = cookbook
         self.force = force
-        self.no_deps = no_deps
         self.missing_files = missing_files
         shell.DRY_RUN = dry_run
 
-    def start_cooking(self):
+    def start_cooking(self, recipes):
```
ylatuya: This could be simplified just by modifing the list of steps to perform

cerbero/commands/package.py
turran: Actually we both assumed it wrong, check https://docs.python.org/2/library/argparse.html
The store_true is like store but also sets a default value, to True in this case. So or we either use everywhere just store where a default value is set, or store_true/store_false and remove the default option.

cerbero/build/cookbook.py
turran: Didn't know about the spurious message. This is to fix the case where using the -f when building or a buildone the steps will not be added to the list of steps done (as those were already done), just an optimization.

cerbero/build/fridge.py
turran: Ok, already commented but to clarify: If we use a Mixin, it means that a recipe can overwrite the binary methods, it makes sense for the classic build methods but not for the package/upload/download/extract steps, those must be immutable

cerbero/commands/package.py
ylatuya: The old behaviour was correct, except that it was saving devel in a variable named no_devel.
So when nothing was set, no_devel = False, when -d was passed store_false was storing false in no_devel.
If you change the behaviour to store no_devel in args.no_devel, so by dedault it's no_devel=True, so also need to change action to store_true, so when -d is passed no_devel=True.

PR #23, opened 2016-11-07
Don't override existing packages with the ones from recipes

vguzmanp:
Fix for commit 456671056d4dc9ffa92fb30eec3535c34d01f9e8 @turran


turran:
Seems good, we'll have issues now when a recipe finishes and tries to create the package for such recipe (for later uploading), now it will try to create the main package, not the one from the recipe ... let's merge this until we find a good solution

PR #24, opened 2016-11-13
Lon 918

ylatuya:


recipes/mono/0001-System.ServiceModel-Fix-DuplexClient-initialization-.patch
```diff
@@ -0,0 +1,60 @@
...
+ 		}
+ 
+ 		internal ClientBase (ChannelFactory<TChannel> factory)
+@@ -150,6 +150,12 @@ namespace System.ServiceModel
+ 			ChannelFactory = factory;
+ 		}
+ 
++		internal virtual void Initialize (InstanceContext instance, ServiceEndpoint endpoint)
++		{
++			// FIXME: use instance
```
vguzmanp: So... why can't we use it like in the `Duplex` one?

recipes/gtk-sharp/gtk-sharp-2.0.recipe
```diff
@@ -3,13 +3,12 @@ import shutil
...
 class Recipe(recipe.Recipe):
     name = 'gtk-sharp'
-    version = '2.12.10'
+    version = '2.12.42'
     licenses = [License.LGPL]
     stype = SourceType.TARBALL
     deps = ['gtk+', 'libglade', 'mono']
-    url = 'http://ftp.gnome.org/pub/gnome/sources/gtk-sharp/2.12/gtk-sharp-2.12.10.tar.gz'
-    patches = ['0001-fix_glib_imports.patch']
-    strip = '1'
```
vguzmanp: lost the strip?

recipes/mono/0001-System.ServiceModel-Fix-DuplexClient-initialization-.patch
ylatuya: What do you mean ?

vguzmanp: In [here](https://github.com/fluendo/cerbero/pull/24/files/ea8e91bfd9fa3662b41eabcb99107e7993ee92fe#diff-8611ad6e623b7f78ecacb28232266c16R52) you call the `DuplexChannelFactory`and you use `instance`, but here it's as a FIXME, can't we do it the same way?

ylatuya: I have just duplicated the comment, like in the rest of constructors:
https://github.com/ylatuya/mono/blob/33cd0a6185ae231c3601e5f457f18c309e5d9dd4/mcs/class/System.ServiceModel/System.ServiceModel/ClientBase.cs#L147

vguzmanp: :+1:

PR #25, opened 2016-11-16
Update to Mono 4.6 in mono-windows recipe

vguzmanp:


recipes/mono/mono-windows.recipe
```diff
@@ -25,9 +25,10 @@ class Recipe(recipe.Recipe):
         'bin/zlib1.dll',
         'etc/mono',
         'lib/mono/gac',
-        'lib/mono/2.0',
-        'lib/mono/4.0',
+        'lib/mono/2.0-api',
+        'lib/mono/4.0-api',
         'lib/mono/4.5',
```
ylatuya: lib/mono/4.5 still exists?

vguzmanp: It was in the buildbot's folder `C:\Program Files\Mono\lib\mono`. Let me check if it's in a fresh install too.

vguzmanp: Yes, it still exists: 

> The 4.5 folder contains the actual assemblies that are used during runtime. We’re considering these the latest assemblies from now on.
> http://www.mono-project.com/docs/about-mono/releases/4.4.0/

Actually in a fresh install there is also the `4.0` folder, but I'm not sure we need that one.

PR #27, opened 2016-11-28
Vas 34 - Use a Burn Packager for riftanalyst

vguzmanp:


cerbero/packages/wix.py
```diff
@@ -270,13 +271,25 @@ def __init__(self, config, package):
...
+    def copy(self, output_dir):
         config_out_path = os.path.join(output_dir,
                 os.path.basename(self.wix_config))
         shutil.copy(self.config_path, os.path.join(output_dir,
                     os.path.basename(self.wix_config)))
-        replacements = {
+        return config_out_path
+
+    def replace(self, config_out_path):
+        shell.replace(config_out_path, self.replacements)
```
vmartos: One method for only one line. You can call shell.replace in the write method

cerbero/packages/wix.py
```diff
@@ -571,3 +582,66 @@ def _add_start_menu_shortcuts(self):
...
+        # for path, var in [(self.BANNER_BMP, 'BannerBmp'),
+        #         (self.DIALOG_BMP, 'DialogBmp'),
+        #         (self.LICENSE_RTF, 'LicenseRtf')]:
+        #     path = self.package.relative_path(path)
+        #     if self._with_wine:
+        #         path = to_winepath(path)
+        #     if os.path.exists(path):
+        #         etree.SubElement(self.product, 'WixVariable',
+        #                 Id='WixUI%s' % var, Value=path)
+        # # Icon
```
vmartos: Is commented because in this type of installer this is not needed?

cerbero/packages/wix_packager.py
```diff
@@ -297,13 +356,16 @@ def __init__(self, wix_prefix, with_wine, extra=''):
...
             self.options['q'] = ''
 
-    def compile(self, objects, msi_file, output_dir, merge_module=False):
+    def compile(self, objects, msi_file, output_dir, merge_module=False, ext=None):
         self.options['objects'] = ' '.join(objects)
         self.options['msi'] = msi_file
-        if merge_module:
-            self.options['ext'] = 'msm'
+        if ext:
+            self.options['ext'] = ext
```
vmartos: what is ext? if it's extension Can you add that rename the variable with that to be more clear.

cerbero/packages/wix.py
vguzmanp: Although now that I'm looking at it, I probably don't need the two methods, as I already have the init_replacements method.

cerbero/packages/wix.py
vguzmanp: It was commented because I wasn't sure how the UI was going to be. Now that the UI is almost finished, this can go away.

vmartos:
👍

PR #29, opened 2016-12-09
Inf 72

turran:


cerbero/config.py
```diff
@@ -71,6 +72,17 @@ def __getattr__(self, name):
         except Exception:
             raise AttributeError("%s is not a known variant" % name)
 
+    def __iter__(self):
+        for attr, value in self.__dict__.iteritems():
```
ylatuya: Isn't this iterating over all fields, even the private fields __disabled_variants and __enabled_variants ?

cerbero/config.py
```diff
@@ -234,6 +246,7 @@ def get_env(self, prefix, libdir, py_prefix):
         monopath = self._join_path (os.path.join(libdir, 'mono', '4.5'),
             os.path.join(libdir, 'mono', '4.5', 'Facades'))
 
+        variants = ':'.join(self.variants.get_enabled())
```
ylatuya: I think All variants should be listed in their current state, no just enabled ones.

cerbero/config.py
```diff
@@ -372,6 +373,16 @@ def cross_compiling(self):
                 self.target_arch != self.arch or \
                 self.target_distro_version != self.distro_version
 
+    def get_md5(self):
+        env = {}
+        md5 = hashlib.md5()
+        for e,v in self.env.iteritems():
```
ylatuya: items should be sorted just to ensure order is also kept.

cerbero/config.py
```diff
@@ -372,6 +373,16 @@ def cross_compiling(self):
                 self.target_arch != self.arch or \
                 self.target_distro_version != self.distro_version
 
+    def get_md5(self):
+        env = {}
```
ylatuya: env is never used.

cerbero/build/cookbook.py
```diff
@@ -390,6 +394,12 @@ def _load_recipes(self):
                     st.touch()
                 else:
                     self.reset_recipe_status(recipe.name)
+            # check that we use the same config md5
+            # Use getattr as config_hash was added later
+            saved_config_hash = getattr(st, 'config_hash', 0)
```
ylatuya: I don't recall how deserialization works with new fields, but this could be avoided if config_hash was a class attribute and therefore always initialized to a sane value, same for the other ones.

cerbero/config.py
turran: Yes

cerbero/build/cookbook.py
turran: Honestly I have no idea. I just followed what was there on line 387. I'm just using the same approach as the "file_hash" attribute

PR #38, opened 2017-02-03
Create a distribution zip

marecki315:
Added option for creating z zip distribution. It works similar to
creating a distribution tarball.


cerbero/packages/distzip.py
```diff
@@ -0,0 +1,95 @@
...
+                                         devel_files, force, package_prefix)
+            filenames.append(devel)
+        return filenames
+
+    def _get_name(self, package_type, ext='zip'):
+        return "%s%s-%s-%s-%s%s.%s" % (self.package_prefix, self.package.name,
+                self.config.target_platform, self.config.target_arch,
+                self.package.version, package_type, ext)
+
+    def _create_zip(self, output_dir, package_type, files, force,
```
ylatuya: Instead of creating a new class that duplicates all the logic and implements _create_zip, you could probably have a single DistArchive that can be configured in the constructor to use either zip or gzip (or whatever other archive compression format we need inthe future)

cerbero/commands/package.py
```diff
@@ -41,6 +42,9 @@ def __init__(self):
             ArgparseArgument('-t', '--tarball', action='store_true',
                 default=False,
                 help=_('Creates a tarball instead of a native package')),
+            ArgparseArgument('-z', '--zip', action='store_true',
```
ylatuya: Previously there were only 2 choices, either normal package or tarball, so the --tarball switch was enough. If you add another switch for zip, you could endup with --tarball --zip and generate only a tarball. I would use a "choice" option:
 ArgparseArgument('-t', '--type', type=str, choices=['native', 'tarball', 'zip'], default='native'))
These cahnge is not backwards comatible and we will have to fix the buildbots that uses it too.

cerbero/commands/package.py
```diff
@@ -38,9 +39,9 @@ def __init__(self):
                              help=_('name of the package to create')),
             ArgparseArgument('-o', '--output-dir', default='.',
                              help=_('Output directory for the tarball file')),
-            ArgparseArgument('-t', '--tarball', action='store_true',
-                default=False,
-                help=_('Creates a tarball instead of a native package')),
+            ArgparseArgument('-t', '--type', type=str, default='native',
+                choices=['native', ArchiveType.TARBALL , ArchiveType.ZIP],
+                help=_('Choose a package type, navtive or archive')),
```
ylatuya: typo "navtive"

PR #39, opened 2017-02-06
Vas 48

ylatuya:


cerbero/utils/git.py
```diff
@@ -155,7 +155,8 @@ def local_checkout(git_dir, local_git_dir, commit):
     shell.call('%s branch build' % GIT, local_git_dir, fail=False)
     shell.call('%s checkout build' % GIT, local_git_dir)
     shell.call('%s reset --hard %s' % (GIT, commit), local_git_dir)
-    return shell.call('%s clone %s -b build .' % (GIT, local_git_dir), git_dir)
+    shell.call('%s clone %s -b build .' % (GIT, local_git_dir), git_dir)
+    shell.call('%s submodule update --init --recursive' % (GIT), git_dir)
```
vmartos: Now it doesn't return the result of the last shell.Call is that ok?

recipes/sparkle/sparkle.recipe
```diff
@@ -0,0 +1,55 @@
...
+            raise InvalidRecipeError ()
+        self.files_nuget.append('share/nuget/Sparkle.%s.nupkg' % self.version)
+        # ld does not understand -isysroot so use cc instead for linking
+        self.new_env['LD'] = 'cc'
+
+    @modify_environment
+    def compile(self):
+        shell.call('xcodebuild -configuration Release -target Sparkle',
+                os.path.join(self.build_dir, 'Sparkle/3rdparty/Sparkle'))
+        shell.call("xbuild Sparkle.sln /p:Configuration=Release", self.build_dir)
```
vmartos: this recipe only builds on MacOS but just to let you know. I Had problems in windows because the option /p wasn't accepted -_-.Is needed to put the entire word in the options, Ex: /property:Configuration=Release.

recipes/winsparkle.recipe
```diff
@@ -0,0 +1,31 @@
...
+class Recipe(recipe.Recipe):
+    name = 'WinSparkle'
+    version = '0.5.3'
+    licenses = [License.MIT]
+    stype = SourceType.TARBALL
+    btype = BuildType.CUSTOM
+    url = "https://github.com/vslavik/winsparkle/releases/download/v0.5.3/WinSparkle-0.5.3.zip"
+
+    files_lib = [
+            "bin/libsparkle.dll"
```
vmartos: This works? files_lib is for files that should be installed in prefix/lib

recipes/xamarin.mac.recipe
```diff
@@ -0,0 +1,29 @@
...
+    files_cli = [
+        'lib/mono/4.5/Xamarin.Mac.dll',
+    ]
+
+    def prepare(self):
+        if self.config.platform != Platform.DARWIN:
+            raise InvalidRecipeError()
+
+    def install(self):
+        xamarinmac = "/Library/Frameworks/Xamarin.Mac.framework/Versions/Current/"
```
vmartos: Shouldn't this alert to install Xamarin.Mac.Framework in the prepare method if it doesn't exists?

cerbero/utils/git.py
ylatuya: the return type is never used and it was incorrectly returned

recipes/xamarin.mac.recipe
ylatuya: IT's installed by default with XS

recipes/sparkle/sparkle.recipe
ylatuya: weird...

recipes/winsparkle.recipe
ylatuya: files_libs with a final 's'

PR #47, opened 2017-04-05
Inf 81

vguzmanp:


cerbero/build/cookbook.py
```diff
@@ -112,6 +112,7 @@ def set_config(self, config):
         @type config: L{cerbero.config.Config}
         '''
         self._config = config
+        config.cookbook = self
```
vmartos: why this change on the cookbook is needed? I mean why config should have an instance of the cookbook?

vguzmanp: It's because of [this](https://github.com/fluendo/fluendo-cerbero/pull/93/files#diff-29db32840edad6b8c349544da19b8f3fR735)

 That was creating a new CookBook, which in turn was iterating over all available recipes and creating the classes again. After this, the `super` calls never worked, because the new classes weren't descendants of the old ones (and viceversa).

Instead of creating a new cookbook, we get the one already stored in the config, and everything works again.

vmartos: 👍

PR #51, opened 2017-04-24
Don't change the install_name if it's a symlink

vguzmanp:
It will already do it in the original file


cerbero/build/fridge.py
```diff
@@ -100,7 +100,8 @@ def extract_binary(self, recipe):
                         extracted_object = os.path.join(self.config.prefix, member.name)
                         relocator = OSXRelocator(self.config.prefix, self.config.prefix, True)
                         # When extracting, we change the install_name of the library to match the path
-                        relocator.change_id(extracted_object, extracted_object)
+                        if not os.path.islink(extracted_object):
+                            relocator.change_id(extracted_object, extracted_object)
```
ylatuya: reuse the same instance of the relocator.

PR #52, opened 2017-04-26
Append ssl libraries

aameijide:


recipes/openssl.recipe
```diff
@@ -91,6 +91,9 @@ class Recipe(recipe.Recipe):
             self.files_devel.append('lib/libcrypto.a')
             self.files_devel.append('lib/libcrypto.dll.a')
             self.files_devel.append('lib/libcrypto.la')
+			self.files_devel.append('lib/libssl.a')
```
ylatuya: extra spaces

aameijide: It seems is git hub, in my computer I see correctly.

PR #53, opened 2017-04-26
Fix recipes files - LON-1049

vguzmanp:
LON-1049 is already merged with the changes in the relocation with symlinks, but these changes are related.


recipes/mono.recipe
```diff
@@ -70,3 +72,263 @@ class Recipe(recipe.Recipe):
         shell.call ('rm -rf lib/mono/4.5/nunit*', self.config.prefix)
         shell.call ('rm -rf lib/mono/gac/*/2.4.8.0__96d09a1eb7f44a77',
                 self.config.prefix)
+
+    # This is the list of installed files that we are not listing in the recipe, in case we ever need them
```
ylatuya: Is this really needed ?

vguzmanp: No, but after a couple times of `X not found -> search the recipes, none is mentioning X -> try the files-missing until you find the right recipe -> add it to the files`, I decided that it will be easier to search these in the future

PR #56, opened 2017-05-11
Introduce global_build_dir

dabrain34:
Allow to change the global build dir where cerbero
creates the build artifacts.

### DIFF
--- a/cerbero/config.py
+++ b/cerbero/config.py
@@ -93,7 +93,7 @@ class Config (object):
                    'distro', 'target_distro', 'environ_dir', 'cache_file',
                    'toolchain_prefix', 'distro_version',
                    'target_distro_version', 'allow_system_libs',
-                   'packages_dir', 'py_prefix',
+                   'packages_dir', 'py_prefix', 'extra_properties',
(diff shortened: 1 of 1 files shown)


ylatuya:
Like we discussed I don't see the point of this new variable in cerbero itself, since it's unused by cerbero and cerbero has already a mechanism for that, which is home_dir.

Instead, to pass external properties I would add a new property field called extra_properties, a dicctionary in which we add random key:values for custom use in extensions like fluendo-cerbero.

For example, in your cerbero.cbc you could then have:
extra_properties =  {'global_build_dir': '/path'}

segfaultxavi:
I agree with ylatuya, Cerbero only needs to know about home_dir. If we use Cerbero as part of a greater buildsystem, that is our problem, not Cerbero's.
I am sorry I did not understand before these changes.

dabrain34:
So ylatuya you'll advise to add this to cerbero/config.py and then use it in fluendo-cerbero right ?

dabrain34:
This pull request needs https://github.com/fluendo/cerbero/pull/59

PR #59, opened 2017-05-15
Remove reset flag in _parse method

dabrain34:
No need for reset flag anymore as it should not be
used in load_main_config anymore.


segfaultxavi:
I leave the review to Andoni, I do not know what is going on here :P

PR #61, opened 2017-05-18
Add an optional libname to genlib

vguzmanp:
It will replace it in the .def file

### DIFF
--- a/cerbero/build/recipe.py
+++ b/cerbero/build/recipe.py
@@ -107,6 +107,8 @@ class Recipe(FilesProvider):
     @type deps: list
     @cvar platform_deps: platform conditional depencies
     @type platform_deps: dict
+    @cvar library_name_override: library names to override
+    @type library_name_override dict
     @cvar runtime_dep: runtime dep common to all recipes
     @type runtime_dep: bool
     '''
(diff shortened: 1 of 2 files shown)


cerbero/build/recipe.py
```diff
@@ -232,7 +233,8 @@ def gen_library_file(self, output_dir=None):
                 implib = genlib.create(
                     os.path.join(self.config.prefix, dllpath),
                     self.config.target_arch,
-                    os.path.join(self.config.prefix, 'lib'))
+                    os.path.join(self.config.prefix, 'lib'),
+                    self.library_name_override)
```
ylatuya: This is not going to work, it's iterating over a list filenames, and library_name_override can provide a single override. It should be a boolean that uses the filename instaed of the file path, but since it hard to configure it for a single library it's probably better to just override the gen_library_file int the recipe that needs the special handling

cerbero/build/recipe.py
```diff
@@ -123,6 +123,7 @@ class Recipe(FilesProvider):
     deps = None
     platform_deps = None
     force = False
+    library_name_override = None
```
ylatuya: If you add a new variable make sure you document it in the class as the rest of them

cerbero/build/recipe.py
vguzmanp: You're right.

But instead of overriding the method, what about using a dict? If the key `dllpath` exists, we send the value, else we send None.

PR #64, opened 2017-06-06
Several recipes in one file.

dabrain34:
Cerbero can now detect several recipes in a recipe file
with prefix Recipe...
Can be useful when declaring a demo recipe class.

### DIFF
--- a/cerbero/build/cookbook.py
+++ b/cerbero/build/cookbook.py
@@ -420,24 +420,19 @@ class CookBook (object):
             # which can contain private classes to extend cerbero's recipes
             # and reuse them in our private repository
             try:
-                recipe = self._load_recipe_from_file(f, custom)
+                recipes_from_file = self._load_recipes_from_file(f, custom)
             except RecipeNotFoundError:
                 m.warning(_("Could not found a valid recipe in %s") % f)
-            if recipe is None:
+            if recipes_from_file is None:
                 continue
-            recipes[recipe.name] = recipe
+            for name in recipes_from_file:
+                recipes[name] = recipes_from_file[name]
         return recipes
 
-    def _load_recipe_from_file(self, filepath, custom=None):
+    def _load_recipes_from_file(self, filepath, custom=None):
+        recipes = {}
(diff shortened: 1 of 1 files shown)


cerbero/build/cookbook.py
```diff
@@ -420,21 +420,24 @@ def _load_recipes_from_dir(self, repo):
...
                 continue
-            recipes[recipe.name] = recipe
+            for name in recipes_from_file:
+                recipes[name] = recipes_from_file[name]
         return recipes
 
-    def _load_recipe_from_file(self, filepath, custom=None):
+    def _load_recipes_from_file(self, filepath, custom=None):
+        recipes = {}
         mod_name, file_ext = os.path.splitext(os.path.split(filepath)[-1])
```
ylatuya: To support correctly multiple recipes, you should move the parsing of the file to the beginning of the function, here, and iterate from here in all Recipe* you find in that dicctionary

dabrain34: I'm not sure to understand the difference with what i have changed...Do you want me to do the loop earlier as it was done before ?

cerbero/build/cookbook.py
```diff
@@ -451,20 +454,23 @@ def _load_recipe_from_file(self, filepath, custom=None):
...
                         conf.prefix = os.path.join(self._config.prefix, c)
-                r = d['Recipe'](conf)
-                r.__file__ = os.path.abspath(filepath)
-                self._config.arch_config[c].do_setup_env()
-                r.prepare()
-                if self._config.target_arch == Architecture.UNIVERSAL:
-                    recipe.add_recipe(r)
-                else:
-                    return r
+                for key in d:
```
ylatuya: for key in [x for x in d if x.startswith('Recipe')]

cerbero/build/cookbook.py
turran: I suppose he meant to keep the _load_recipe_from_file parsing method, create a new one named _load_recipes_from_file that will iterate over all the Recipes found on the file and call _load_recipe_from_file there.

PR #65, opened 2017-06-06
Replace the prefix with all the possible paths

vguzmanp:
Windows, unix, wine...

Replace the prefix in text files in bin too


vguzmanp:
Totally agreed...

PR #68, opened 2017-06-16
Optional list of 'Replaces' and 'Provides' while packaging

marecki315:


cerbero/packages/package.py
```diff
@@ -111,6 +111,8 @@ class PackageBase(object):
     resources_postinstall = 'postinstall'
     resources_postremove = 'postremove'
     conflicts = ''
+    replaces = ''
```
ylatuya: Since those are debian specific, prepend debian_ to the variable name. Is there an equivalent for RPM ?

cerbero/packages/package.py
```diff
@@ -111,6 +111,8 @@ class PackageBase(object):
...
+    provides = ''
```
ylatuya: The new variables should be documented as its done for the rest of variables like in https://github.com/fluendo/cerbero/blob/147c8dbd058dc6836d0d813258edcf7e338a4bdd/cerbero/packages/package.py#L82

cerbero/packages/package.py
marecki315: Yes, those are debian specific. There is a equivalent in RPM. 'conflicts' is used also in RPM. Names of packages might differ for debian and RPM so maybe we should have separated set of variables for both?

ylatuya: ok!

PR #69, opened 2017-06-19
debian: Replace %(files)s template inside postinstall and postremove scripts

marecki315:
Same thing was already done for those scripts while packaging
rpm (commit 6320598c58266d48e05f6b4b2b57ce3f89517ea7)

### DIFF
--- a/cerbero/packages/debian.py
+++ b/cerbero/packages/debian.py
@@ -201,12 +201,19 @@ class DebianPackager(LinuxPackager):
         os.mkdir(os.path.join(packagedir, 'source'))
         m.action(_('Creating debian package structure at %s for package %s') %
                 (srcdir, self.package.name))
+        files_list = self.files_list(PackageType.RUNTIME)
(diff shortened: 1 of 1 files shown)


cerbero/packages/debian.py
```diff
@@ -201,12 +201,19 @@ def create_tree(self, tmpdir):
         os.mkdir(os.path.join(packagedir, 'source'))
         m.action(_('Creating debian package structure at %s for package %s') %
                 (srcdir, self.package.name))
+        files_list = self.files_list(PackageType.RUNTIME)
+        files = '\n'.join([os.path.join(self.install_dir + '/' + f) \
+            for f in files_list])
         if os.path.exists(self.package.resources_postinstall):
-            shutil.copy(os.path.join(self.package.resources_postinstall),
-                        os.path.join(packagedir, 'postinst'))
+            script = open(self.package.resources_postinstall).read()
```
ylatuya: For consistency I would use shell.replace, keeping the old file copy +  shell.replace

PR #75, opened 2017-08-08
update mono version to 5.0.1.1

vmartos:
- update mono windows recipe
- add mono-osx recipe

### DIFF
--- a/cerbero/tools/osxrelocator.py
+++ b/cerbero/tools/osxrelocator.py
@@ -76,6 +76,14 @@ class OSXRelocator(object):
                                                object_file)
                 shell.call(cmd, fail=False)
 
+    def change_lib_path(self, object_file, old_path, new_path):
+        for lib in self.list_shared_libraries(object_file):
+            if old_path in lib:
+                new_path = lib.replace(old_path, new_path)
+                cmd = '%s -change "%s" "%s" "%s"' % (INT_CMD, lib, new_path,
+                                               object_file)
+                shell.call(cmd, fail=True)
+
     def parse_dir(self, dir_path, filters=None):
         for dirpath, dirnames, filenames in os.walk(dir_path):
             for f in filenames:
--- a/recipes/mono.recipe
+++ b/recipes/mono.recipe
@@ -3,10 +3,10 @@ from cerbero.utils import shell
 
 class Recipe(recipe.Recipe):
     name = 'mono'
-    version = '4.6.0'
+    version = '5.0.1.1'
     licenses = [License.LGPL]
     stype = SourceType.TARBALL
-    url = 'http://download.mono-project.com/sources/mono/mono-%(version)s.245.tar.bz2'
+    url = 'http://download.mono-project.com/sources/mono/mono-%(version)s.tar.bz2'
     configure_options = ' --with-gc=included --with-sgen=yes '\
                         ' --with-preview=yes' \
                         ' --disable-dtrace ' \
@@ -30,18 +30,10 @@ class Recipe(recipe.Recipe):
         'lib/pkgconfig/mono.pc',
     ]
 
-    patches = [
-        'mono/0001-System.ServiceModel-Fix-DuplexClient-initialization-.patch',
-        'mono/0002-System.ServiceMode-Fix-XML-parsing-of-EndpointAddres.patch',
-        'mono/0003-System.ServiceModel.Discovery-Fix-deadlock-receiving.patch',
-        'mono/0004-System.ServiceModel.Discovery-Fix-exception-when-Loc.patch',
-        'mono/0005-System.ServiceModel.Discovery-Add-integration-test-f.patch',
-        'mono/0006-System.ServiceMode-Create-unit-test-for-46971.patch',
-        'mono/0007-System.ServiceModel.Discovery-Fix-messages-validatio.patch',
-    ]
-
     def prepare(self):
         if self.config.target_platform == self.config.platform and self.config.target_platform == Platform.WINDOWS:
+            raise InvalidRecipeError()
+        if self.config.target_platform == self.config.platform and self.config.target_platform == Platform.DARWIN:
             raise InvalidRecipeError()
         if self.config.target_platform == Platform.WINDOWS:
             self.configure_options += ' --without-x --enable-win32-dllmain=yes --with-libgc-threads=win32'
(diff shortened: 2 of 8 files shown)


cerbero/tools/osxrelocator.py
```diff
@@ -76,6 +76,14 @@ def change_libs_path(self, object_file):
                                                object_file)
                 shell.call(cmd, fail=False)
 
+    def change_lib_path(self, object_file, old_path, new_path):
+        for lib in self.list_shared_libraries(object_file):
+            if old_path in lib:
+                new_path = lib.replace(old_path, new_path)
+                cmd = '%s -change "%s" "%s" "%s"' % (INT_CMD, lib, new_path,
+                                               object_file)
+                shell.call(cmd, fail=False)
```
vguzmanp: Shouldn't this `fail=True`? If the path is not changed, it won't work properly after unfreezing

recipes/mono.recipe
```diff
@@ -30,19 +30,11 @@ class Recipe(recipe.Recipe):
...
     ]
 
-    patches = [
-        'mono/0001-System.ServiceModel-Fix-DuplexClient-initialization-.patch',
-        'mono/0002-System.ServiceMode-Fix-XML-parsing-of-EndpointAddres.patch',
-        'mono/0003-System.ServiceModel.Discovery-Fix-deadlock-receiving.patch',
-        'mono/0004-System.ServiceModel.Discovery-Fix-exception-when-Loc.patch',
-        'mono/0005-System.ServiceModel.Discovery-Add-integration-test-f.patch',
-        'mono/0006-System.ServiceMode-Create-unit-test-for-46971.patch',
-        'mono/0007-System.ServiceModel.Discovery-Fix-messages-validatio.patch',
```
vguzmanp: 🎉

recipes/mono.recipe
```diff
@@ -30,19 +30,11 @@ class Recipe(recipe.Recipe):
...
-    ]
-
     def prepare(self):
         if self.config.target_platform == self.config.platform and self.config.target_platform == Platform.WINDOWS:
             raise InvalidRecipeError()
+        if self.config.target_platform == self.config.platform and self.config.target_platform == Platform.DARWIN:
+            raise InvalidRecipeError()
```
vguzmanp: So... This only works on linux or when crosscompiling?

vguzmanp: Add a message telling what recipe should be used

recipes/mono/mono-osx.recipe
```diff
@@ -0,0 +1,149 @@
...
+
+        monosgen_link = os.path.join(self.config.prefix, 'lib', 'libmonosgen-2.0.dylib')
+        mono_link = os.path.join(self.config.prefix, 'lib', 'libmono-2.0.dylib')
+
+        shutil.move(os.path.join(self.config.prefix, 'bin', 'mono-sgen64'),
+                    os.path.join(self.config.prefix, 'bin', 'mono-sgen'))
+        shutil.copy(os.path.join(self.config.prefix, 'bin', 'mono-sgen'),
+                    os.path.join(self.config.prefix, 'bin', 'mono'))
+        if(os.path.exists(mono_link)):
+            os.remove(mono_link)
```
vguzmanp: It's better to do it like this:
```
try:
    os.unlink(mono_link)
except IOError as e:
    pass
```
`os.path.exists` has some weird behaviours with symlinks, and you also protect against the link being modified while you check.

recipes/mono/mono-osx.recipe
```diff
@@ -0,0 +1,149 @@
...
+        if(os.path.exists(monosgen_link)):
+            os.remove(monosgen_link)
+        os.symlink(os.path.join(self.config.prefix, 'lib', 'libmonosgen-2.0.1.dylib'),
+                   monosgen_link)
+        os.symlink(monosgen_link,
+                   mono_link)
+
+    def inplace_change(self, filename, old_string, new_string):
```
vguzmanp: This is common enough to be in utils or something like that.

vguzmanp: Also it's good etiquette and useful for the future to add a comment with the stack overflow post where you found it

recipes/mono/mono-osx.recipe
```diff
@@ -0,0 +1,149 @@
...
+            prefix += '/'
+        shell.replace(config, {prefix: '',
+            'libintl.dylib': 'libintl.8.dylib',
+            'libintl.so': 'libintl.8.so',
+            'libgdiplus.dylib': 'libgdiplus.0.dylib',
+            'libgdiplus.so': 'libgdiplus.so.0',
+            })
+        shell.call ('rm -rf lib/mono/2.0/nunit*', self.config.prefix)
+        shell.call ('rm -rf lib/mono/4.5/nunit*', self.config.prefix)
+        shell.call ('rm -rf lib/mono/gac/*/2.4.8.0__96d09a1eb7f44a77',
```
vguzmanp: These are copied from the old mono recipe, right? Shouldn't they be updated to 5.0?

recipes/mono/mono-osx.recipe
```diff
@@ -0,0 +1,149 @@
...
+        relocator.change_id(os.path.join(self.config.prefix, 'lib','libmonosgen-2.0.1.dylib'))
+        relocator.change_id(os.path.join(self.config.prefix, 'lib','libMonoPosixHelper.dylib'))
+        relocator.change_lib_path(os.path.join(self.config.prefix, 'lib','libmonosgen-2.0.1.dylib'),
+            '/usr/lib/libiconv.2.dylib', '@rpath/lib/libiconv.2.dylib')
+        relocator.change_lib_path(os.path.join(self.config.prefix, 'lib','libMonoPosixHelper.dylib'),
+            '/usr/lib/libiconv.2.dylib', '@rpath/lib/libiconv.2.dylib')
+        relocator.change_lib_path(os.path.join(self.config.prefix, 'bin','mono'),
+            '/usr/lib/libiconv.2.dylib', '@rpath/lib/libiconv.2.dylib')
+        relocator.change_lib_path(os.path.join(self.config.prefix, 'bin','mono-sgen'),
+            '/usr/lib/libiconv.2.dylib', '@rpath/lib/libiconv.2.dylib')
```
vguzmanp: reuse `/usr/lib/libiconv.2.dylib` and `@rpath/lib/libiconv.2.dylib` in variables

recipes/mono.recipe
vmartos: yes

vmartos: since mono-osx recipe have the same name it will use this recipe instead of mono.recipe is the same procedure as windows

recipes/mono/mono-osx.recipe
vmartos: no, mono 5.0 still use 4.5 and 4.6 framework versions

PR #80, opened 2017-11-23
Inf 43

turran:


cerbero/packages/wix_packager.py
```diff
@@ -104,11 +105,17 @@ def create_merge_module(self, output_dir, package_type, force, version,
 
         # Clean up
         if not keep_temp:
-            os.remove(sources[0])
+            source_0 = sources[0]
```
ylatuya: I think it's better to keep a copy of wixobjs and sources (see line 98)

cerbero/packages/wix_packager.py
```diff
@@ -121,8 +128,8 @@ def _is_app(self):
             isinstance(self.package, AppExtensionPackage)
 
     def _package_name(self, version):
-        return "%s-%s-%s" % (self.package.name, version,
-                             self.config.target_arch)
+        return "%s-%s-%s" % (self.package.name, self.config.target_arch,
```
ylatuya: OK for that change, but it's probably going to break other things like the release upload scripts

cerbero/packages/wix_packager.py
```diff
@@ -236,11 +243,17 @@ def _create_msi(self, config_path):
 
         # Clean up
         if not self.keep_temp:
-            os.remove(sources[0])
+            source_0 = sources[0]
```
ylatuya: same here, keep a copy of wixobjs and sources

cerbero/utils/__init__.py
```diff
@@ -105,6 +105,13 @@ def to_winepath(path):
         return path
 
 
+def from_winepath(path):
```
ylatuya: this shouldn't be needed keeping the original list of files

recipes/mono.recipe
```diff
@@ -12,7 +13,11 @@ class Recipe(recipe.Recipe):
                         ' --disable-dtrace ' \
                         ' --with-mcs-docs=no '
     config_sh = 'sh autogen.sh'
-    files_bins = ['mono-sgen', 'mono', 'xbuild']
+    platform_files_bins = {
+        Platform.DARWIN: ['mono-sgen', 'mono', 'xbuild'],
```
ylatuya: if xbuild is the only exception, use platform_files_bins only for this one and keep the others in files_bins

cerbero/packages/wix_packager.py
turran: The reasoning was to have the same name schema as other packages, I'll modify the upload script then too, wasn't there when I did this patch :)

turran: Looking at the code, the change was done on the merge module packager, there's no need to change the upload script

PR #81, opened 2017-11-30
Fix gst-plugins-good-1.0 build

dabrain34:
Remove an old define from vpx library.
OP-27


cerbero/build/source.py
```diff
@@ -247,6 +248,11 @@ def extract(self):
 
         # checkout the current version
         git.local_checkout(self.build_dir, self.repo_dir, self.commit)
+        # apply the patches only after checkout to avoid failure with patches already applied.
+        for patch in self.patches:
```
ylatuya: Patches is not documented.
patches is not correctly initialized as it's done in Tarball

PR #84, opened 2017-12-18
Add flag to enable/disable gtkdoc when building with autotools

jbouzas:
This is needed so that we can generate the documentation for every plugin.

### DIFF
--- a/cerbero/build/build.py
+++ b/cerbero/build/build.py
@@ -280,6 +280,12 @@ class Autotools (MakefilesBase):
         if use_configure_cache and self.can_use_configure_cache:
             cache = os.path.join(self.config.prefix, '.configure.cache')
             self.config_sh += ' --cache-file=%s' % cache
+        if self.config.variants.doc:
+            if self.use_gtkdoc:
+                self.configure_tpl += ' --enable-gtk-doc'
+        else:
+            if self.use_gtkdoc:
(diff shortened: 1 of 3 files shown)


cerbero/build/build.py
```diff
@@ -221,6 +221,7 @@ class Autotools (MakefilesBase):
     add_host_build_target = True
     can_use_configure_cache = True
     supports_cache_variables = True
+    enable_gtkdoc = False
```
turran: Instead of creating an enable_gtkdoc variable, better do it in two steps:
* Define a variable like you did named use_gtkdoc
* Define a new variant called 'docs', similar to 'debug' or 'static', etc.

cerbero/build/build.py
```diff
@@ -281,6 +282,9 @@ def configure(self):
             cache = os.path.join(self.config.prefix, '.configure.cache')
             self.config_sh += ' --cache-file=%s' % cache
 
+        if self.enable_gtkdoc:
```
turran: Instead of enable_gtkdoc we better do:
```
if self.config.variants.doc:
    if self.use_gtkdoc:
        self.configure_tpl += ' --enable-gtk-doc'
else:
    if self.use_gtkdoc:
        self.configure_tpl += ' --disable-gtk-doc'
endif
```

jbouzas:
@turran I finished addressing all your comments

PR #85, opened 2017-12-20
recipes/libvpx: Adapt patches from cerbero community to be applied to…

aslobodeniuk:
… v1.6.1


aslobodeniuk:
Comparing patches for 1.6.0 with 1.6.1 shows that one of them was applied, and one needs some simple modification (just different start number of line).

Also some unused patches are deleted.

PR #89, opened 2017-12-27
Debian: Add intltool to apt-get list.

aslobodeniuk:
Continues INF-165: build support for Debian 9 (Stretch).


aslobodeniuk:
> This should have been added only for Debian 9 case right? not generic for every debian version
For every debian version. "intltool" package exists in all supported debian versions (by debian project), and "intltool" package really should be in apt-get list because we can't know if it's missing or not by default in different builds of debian.

turran:
Alright, given that for older debian versions it was working I was surprised, ok to go.

PR #90, opened 2017-12-28
Updating default version of gettext-m4, to make

aslobodeniuk:
m4 doesn't fail on OSX 10.13 High Sierra.
(INF-163)


aslobodeniuk:
This commit is not actually finished at all for next reasons:
the changes make the gettext-m4 recipe be the same as
`fluendo-cerbero/recipes-oneplay-transcoder/build-tools/gettext-m4.recipe`

this means I also should:
- ask David Fernandez, why he copied and modified (probably) the recipe into oneplay-transcoder recipes
- remove file from recipes-oneplay-transcoder because it's not needed anymore

PR #91, opened 2017-12-29
recipes/build-tools/m4: add patch for m4 from macports, so it works

aslobodeniuk:
correctly on OSX 10.13 High Sierra

same patch (with difference in start line and paths sometimes)
 also applies for:
- glib
- gnutls
- build-tools/gettext-tools
- build-tools/gettext-m4


aslobodeniuk:
The patch is originally taken from:
https://raw.githubusercontent.com/macports/macports-ports/edf0ee1e2cf/devel/m4/files/secure_snprintf.patch

The meaning is to disable %n for Darwin in general,
same as in patch from gnulib's mailing list:
http://lists.gnu.org/archive/html/bug-gnulib/2017-07/msg00057.html

PR #92, opened 2018-01-02
recipes/build-tools/m4: add patch for m4 from macports, so it works

aslobodeniuk:
correctly on OSX 10.13 High Sierra

same patch (with difference in start line and paths sometimes)
 also applies for:
- glib
- gnutls
- build-tools/gettext-tools
- build-tools/gettext-m4


aslobodeniuk:
this is a backport to oneplay-release, same commit, but different commit's parent

PR #94, opened 2018-01-05
recipes/gnutls : fetch from https instead of ftp

aslobodeniuk:
Curl fails on fetching from gnutls's website by ftp.


aslobodeniuk:
@turran 
but the commit message contains:
"Curl fails on fetching from gnutls's website by ftp."

PR #101, opened 2018-01-11
For crosscompiling to x86 and x86_64 let user just specify the "-m" f…

aslobodeniuk:
…lag.

Reason: to skip installing toolchain for easy cases.
Automake should handle it and give priority to cross-toolchain first,
we just don't force overriding the compiler.

### DIFF
--- a/config/linux.config
+++ b/config/linux.config
@@ -48,13 +48,20 @@ if target_arch in [Architecture.ARM, Architecture.ARMv7]:
 if target_arch_flags is not None:
     arch_flags += ' %s ' % (target_arch_flags)
 
-if host is None and target_arch != arch:
-    host = _host
+host_was_specified_before = 'true'
+if host is None:
+    if target_arch != arch:
+         host = _host
+    host_was_specified_before = 'false'
 
 tools_prefix = ''
 if host is not None:
     tools_prefix = '%s-' % host
-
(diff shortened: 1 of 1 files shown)


dabrain34:
Could be nice to review the commit message.

config/linux.config
```diff
@@ -48,13 +48,20 @@ if target_arch in [Architecture.ARM, Architecture.ARMv7]:
 if target_arch_flags is not None:
     arch_flags += ' %s ' % (target_arch_flags)
 
-if host is None and target_arch != arch:
-    host = _host
+host_was_specified_before = 'true'
```
ylatuya: It can be simplified. Before doing anything with the host variable store if we need to skip the tools prefix:

skip_tools_prefix = host is None and target_arch not in [Architecture.X86, Architecture.X86_64]

config/linux.config
```diff
@@ -48,13 +48,20 @@ if target_arch in [Architecture.ARM, Architecture.ARMv7]:
...
+if host is None:
+    if target_arch != arch:
+         host = _host
+    host_was_specified_before = 'false'
 
 tools_prefix = ''
 if host is not None:
```
ylatuya: And finally here just do:
if host is None and not skip_tools_prefix:

PR #103, opened 2018-01-15
Fix linux cross compilation: x86_64->x86

aslobodeniuk:
Remove toolchain prefix and use only the
gcc -m option to define the target architecture.

### DIFF
--- a/cerbero/config.py
+++ b/cerbero/config.py
@@ -151,6 +151,8 @@ class Config (object):
         self._raw_environ = os.environ.copy()
 
         for config in self.arch_config.values():
+            if config == self:
+                continue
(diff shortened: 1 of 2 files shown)


dabrain34:
Can you review the title fo something like that
linux cross-compile across processors address space

aslobodeniuk:
@dabrain34 
message is fixed, so as title of pull request.

dabrain34:
Fix linux cross compilation: x86_64->x86

Remove toolchain prefix and use only the 
gcc -m option to define the target architecture.

aslobodeniuk:
@dabrain34 done.

PR #106, opened 2018-02-07
Inf 191-Fix gstreamer-sdk msi package creation using linux

dabrain34:


cerbero/packages/wix.py
```diff
@@ -71,21 +71,16 @@ def __init__(self, config, package):
...
         # and should not be a folder.
         if self._with_wine:
-          self.wix_wine_prefix = '/tmp/flu_cwwp'
-          if os.path.exists(self.wix_wine_prefix):
-            os.remove(self.wix_wine_prefix)
-          os.symlink(config.prefix, self.wix_wine_prefix)
-          self.prefix = self.wix_wine_prefix
+          if os.path.exists(get_wix_wine_prefix()):
+            os.remove(get_wix_wine_prefix())
+          os.symlink(config.prefix, get_wix_wine_prefix())
```
ylatuya: In Windows os.symlink creates a full copy of the prefix

cerbero/packages/wix.py
```diff
@@ -71,21 +71,16 @@ def __init__(self, config, package):
...
+          self.prefix = get_wix_wine_prefix()
         else:
           self.prefix = config.prefix
         self.filled = False
         self.id_count = 0
         self.ids = {}
 
-    def __del__(self):
```
ylatuya: This class is longer destructing the symlink created int he constructor

cerbero/packages/wix.py
```diff
@@ -497,16 +498,20 @@ def _add_merge_modules(self):
...
-        for package, required, selected in packages:
-            if package in self.packages_deps:
-                self._add_merge_module(package, required, selected,
+        if not self.package.wix_use_fragment:
+          for package, required, selected in packages:
+              if package in self.packages_deps:
+                  self._add_merge_module(package, required, selected,
                                        required_packages)
 
         # Add a merge module ref for all the packages
```
ylatuya: This comment is no longer valid, since we are not always creating merge modules

cerbero/packages/wix_packager.py
```diff
@@ -40,6 +40,11 @@ def __init__(self, config, package, store):
         self._with_wine = config.platform != Platform.WINDOWS
         self.wix_prefix = get_wix_prefix()
 
+    def __del__(self):
```
ylatuya: You are destructing here a resource created by someone else

cerbero/packages/wix_packager.py
```diff
@@ -151,6 +155,11 @@ def __init__(self, config, package, store):
...
```
ylatuya: Same here

cerbero/utils/__init__.py
```diff
@@ -104,6 +104,8 @@ def to_winepath(path):
         path = 'z:\\%s' % path
         return path
 
+def get_wix_wine_prefix():
+    return '/tmp/flu_ccwp'
```
ylatuya: Use a random generated path, several build in this same machine could be using the same path

recipes/gst-plugins-bad.recipe
```diff
@@ -152,9 +152,6 @@ class Recipe(recipe.Recipe):
         Platform.WINDOWS: [
              'lib/gstreamer-0.10/libgstd3dvideosink%(mext)s',
              'lib/gstreamer-0.10/libgstdirectdrawsink%(mext)s',
-             'lib/gstreamer-0.10/libgstdshowdecwrapper%(mext)s',
```
ylatuya: Why are those removed unconditionally ? On native platform they are compiled correctly

cerbero/utils/__init__.py
dabrain34: Or I could remove the removal part seen that its only a symlink unique to our buildsystem...

dabrain34:
We have to change the place where it is installed. 
Right now its not using the same logic as 1.0 SDK.
It is installing the package in c:\gstreamer-sdk

cerbero/utils/__init__.py
ylatuya: I don't think that will work, you will still have the same issue of several build configuration using the same directory. Temporal directories should never be hardcoded when possible

cerbero/packages/wix.py
ylatuya: This issue is still pending to be fixed correctly.

cerbero/packages/wix.py
ylatuya: IIRC you told me this is a hack for when the paths are too long, and this issue will happen also in a native windows when the prefix root is too long.
What is exactly the problem with long paths? Is it  with the ID's in the msis packaging step?
If that's the issue than the fix must be handled differently because we generate those id's based on the path, so we could compute those id's on a different way.

cerbero/utils/__init__.py
dabrain34: It is not a temporal directory but a symbolic link to the build directory. I located it in /tmp for this reason to be remove by the linux system at least when we reboot the machine. Concerning the "hack", yes it is a hack to workaround the wine/wix limitation. I guess this bug is not happening on windows, or the use of Z folder is hiding it.

cerbero/packages/wix.py
dabrain34: On windows we dont have the possibility to create symlink but i guess using Z:\ as the build directory path, is hiding the bug, windows 10 I think does not own this bug anymore.
Its not a question of IDs but of resources path.

ylatuya: And what is exactly the error? I would like to read to understand the real issue

cerbero/utils/__init__.py
ylatuya: It's a temporary directory that is created and destroyed in the package step, it can't be hardcoded.
If you need a link for shorter paths that is persistent than create it your self, don't do it with cerbero, just link /srv/yourbuild to /the/real/long/path/.

cerbero/packages/wix.py
ylatuya: This one is still pending

cerbero/packages/wix.py
dabrain34: The error is related to the len of the file path. On linux it can /opt/build_blablab or even as my setup is right now in cerbero folder, ie $HOME/DEV/INTERNAL/fluendo-cerbero/build_cerbero. This path can exceed in old limitation of 128 chars or 256 depending on the OS or .NET 
https://github.com/wixtoolset/issues/issues/5314

cerbero/utils/__init__.py
dabrain34: this is a limitation due to wix using .NET4 with wine. I think we should let cerbero perform this, seen that it is in /tmp/

cerbero/packages/wix.py
```diff
@@ -71,21 +71,20 @@ def __init__(self, config, package):
         # wix_wine_prefix is a symbolic link to the regular cerbero prefix
         # and should not be a folder.
         if self._with_wine:
-          self.wix_wine_prefix = '/tmp/flu_cwwp'
-          if os.path.exists(self.wix_wine_prefix):
-            os.remove(self.wix_wine_prefix)
-          os.symlink(config.prefix, self.wix_wine_prefix)
-          self.prefix = self.wix_wine_prefix
+          if os.path.exists(package.wix_wine_prefix):
```
ylatuya: wix_wine_prefix shouldn't be a package variable.
Creating the symlink at this level is not the correct approach, it should be created at the most top level of the packaging step, in MSIPackager, so you ensure that it's created only once at the beginning of the packaging process.
Note that there is alreay a similar approach taken when a package needs stripping, that creates temporary directory and sets the prefix variable in the merge module.
See  https://github.com/fluendo/cerbero/blob/b11acf9bb9abc3b95948ce962ff0c3b2f26f6c6f/cerbero/packages/wix_packager.py#L68
and
https://github.com/fluendo/cerbero/blob/b11acf9bb9abc3b95948ce962ff0c3b2f26f6c6f/cerbero/packages/wix_packager.py#L91

cerbero/utils/__init__.py
```diff
@@ -104,6 +105,8 @@ def to_winepath(path):
         path = 'z:\\%s' % path
         return path
 
+def get_wix_wine_prefix():
+    return '/tmp/flu_%s' % str(uuid.uuid4())[:8]
```
ylatuya: I would rename flu_ to wix_ this temporary directory has nothing specific to fluendo

PR #112, opened 2018-03-06
Add gst1 to default disabled variants

jbouzas:


aslobodeniuk:
but why??
It would be useful to explain the reason in git message.

jbouzas:
This is needed so that we can check for the gst1 variant in the "prepare" step of a recipe. If a *.cbc file does not have the "gst1" variant, python will fail when checking for it. IMO, all variants should be added in one of those lists to know their default value.

PR #113, opened 2018-03-06
Remove build dir during extraction of Git based repo

marecki315:
This will make cerbero apply new or changed patches on already builded
recipes

### DIFF
--- a/cerbero/build/source.py
+++ b/cerbero/build/source.py
@@ -241,25 +241,13 @@ class Git (GitCache):
 
     def extract(self):
         if os.path.exists(self.build_dir):
-            # fix read-only permissions
-            if self.config.platform == Platform.WINDOWS:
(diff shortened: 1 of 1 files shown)


cerbero/build/source.py
```diff
@@ -241,25 +241,27 @@ def __init__(self):
...
-                shell.call('chmod -R +w .git/', self.build_dir, fail=False)
-            try:
-                commit_hash = git.get_hash(self.repo_dir, self.commit)
-                checkout_hash = git.get_hash(self.build_dir, 'HEAD')
-                if commit_hash == checkout_hash:
-                    return False
-            except Exception:
-                pass
-            shutil.rmtree(self.build_dir)
+            if len(self.patches) > 0:
```
ylatuya: It's something we should have been doing for a long time, I would remove the directory even of there are no  patches just to be consistent

PR #115, opened 2018-03-20
Extend the wixobjs array only when using fragment

dabrain34:
Fix f9a5333bbb727bb0ab164c555a82e97cf615dbd3


dabrain34:
This patch fixes the apps generation using Win32/64 builders

dabrain34:
Fix the generation of RiftAnalyst and Longomatch.

By example http://builder.fluendo.lan:8076/#/builders/68

PR #119, opened 2018-04-20
Add a way to add custom variable to Config.wxi

dabrain34:
Add wix_config_custom variable


cerbero/packages/package.py
```diff
@@ -421,6 +421,7 @@ class MetaPackage(PackageBase):
     platform_packages = {}
     sdk_version = '1.0'
     resources_wix_installer = None
+    wix_config_custom = None
```
turran: other vars related to wix are in the form foo_wix_bar, instead of wix_foo_bar, try to use the same variable scheme for this

PR #130, opened 2018-06-14
LMA-321 - Set the version correctly in the couchbase build process

vguzmanp:
It checks the VERSION env var to generate the AssemblyInfo used for versioning, and we generate it using TextTransform (we added the NuGet in our couchbase fork)

### DIFF
--- a/recipes/couchbase-lite/couchbase-lite.recipe
+++ b/recipes/couchbase-lite/couchbase-lite.recipe
@@ -46,6 +46,8 @@ class Recipe(recipe.Recipe):
             self.append_env = {}
             self.new_env =  {'MONO_PATH' : None}
             self.new_env["PATH"] = "/Library/Frameworks/Mono.framework/Commands/:" + os.environ["PATH"]
+        self.short_version = '.'.join(self.version.split('.')[:-1])
+        self.major_version = '.'.join(self.short_version.split('.')[:-1])
 
     def extract(self):
         super(recipe.Recipe, self).extract()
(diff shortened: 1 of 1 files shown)


recipes/couchbase-lite/couchbase-lite.recipe
```diff
@@ -46,16 +46,24 @@ class Recipe(recipe.Recipe):
...
         shell.call('git submodule update --init --recursive', self.build_dir)
 
     @modify_environment
     def compile(self):
+        os.environ["VERSION"] = self.version
+        os.environ["NUGET_VERSION"] = self.short_version
         logging.info('Compiling Couchbase')
         shell.call("nuget restore " \
                    "src/Couchbase.Lite.Net45.sln -source 'https://www.nuget.org/api/v2/'", self.build_dir)
+        shell.call("mono src/packages/Mono.TextTemplating.1.3.1/tools/TextTransform.exe"
```
vmartos: isn't easier to just subst strings in a assemblyinfo like we have in longomatch and riftanalyst? instead of having a dependency on Mono.TextTemplating?

vguzmanp: Yes, but the Couchbase devs are doing this (without the dependency with Mono.TextTemplating because they use VS for windows to build)

I tried to find a way to keep our system as close as theirs, so that when they change something, it applies to us too.

vmartos: 👍

PR #132, opened 2018-06-26
Inf 213

marecki315:


marecki315:
The code modified here was also modified in pull request #131 which was created before, so let's try to merge that one first

marecki315:
Pull request #131 was merged. This branch is rebased sou you can review it.

cerbero/utils/shell.py
```diff
@@ -205,12 +206,15 @@ def download(url, destination=None, recursive=False, check_cert=True, user=None,
...
         cmd += "-r "
-        path = destination
+        path = dst
     else:
-        if destination is not None:
-            cmd += "-O %s " % destination
+        if not dst:
+            dst = ntpath.basename(url)
+        tmp_dst = "%s_tmp" % dst
+        cmd += "-O %s " % tmp_dst
```
dabrain34: wont it be better to have one name for the temp file to avoid having cancelled downloads remaining and in a common path ?

cerbero/utils/shell.py
```diff
@@ -28,6 +28,7 @@
 import glob
 import shutil
 import hashlib
+import ntpath
```
dabrain34: why do you use ntpath ? Shall we use everywhere ? ( ie _get_download_mirror_url)

marecki315: ntpath deals with '\' and also '/' correctly on all platforms. os.path should be enough here. I will change it.

cerbero/utils/shell.py
marecki315: Thought about it but I think it's no necessary. Temporary file will be overwritten during next download try.

dabrain34:
@turran @ylatuya I'd like to to have your advise on this PR, because for me this is too much logic if we have already a cache system which is telling us what has been already performed ( download, extract etc.)
I'd say that having a finer granularity in the state machine of the cache would be enough without adding this intermediary move logic.

turran:
@dabrain34 IMHO the correct approach is not to re-download when an extract fails, but change the download stage to be marked as done only when the whole file is downloaded, i.e not at the beginning of the download process but at the end (after a successful exit status)

cerbero/utils/shell.py
```diff
@@ -267,19 +280,26 @@ def download_curl(url, destination=None, recursive=False, check_cert=True, user=
             cmd += " "
     if not check_cert:
         cmd += "-k "
-    if destination is not None:
-        cmd += "%s -o %s " % (url, destination)
-    else:
-        cmd += "-O %s " % url
+    if not dst:
```
aslobodeniuk: in the code below we expect dst to be == destination. For example at line 288.
Probably this code is actually means "if not destination"

marecki315:
Outdated in current cerbero - closing.

PR #140, opened 2018-08-10
Update bzip2 download link

aslobodeniuk:
Previous is expired


recipes/bzip2.recipe
```diff
@@ -8,7 +8,7 @@ class Recipe(recipe.Recipe):
     version = '1.0.6'
     licenses = [License.BSD_like]
     stype = SourceType.TARBALL
-    url = 'http://bzip.org/1.0.6/bzip2-1.0.6.tar.gz'
```
dabrain34: you can also compose the url with version var such as in taglib or fontconfig etc...

PR #145, opened 2018-08-24
Copy libntdll.a during mingw bootstrap

dabrain34:
In order to avoid issue with ipp, we need this custom
libntdll.a to expose _all_mul and _chkstk


cerbero/bootstrap/windows.py
```diff
@@ -155,6 +156,13 @@ def fix_lib_paths(self):
         for path in [f for f in os.listdir(lib_path) if f.endswith('la')]:
             path = os.path.abspath(os.path.join(lib_path, path))
             shell.replace(path, {orig_sysroot: new_sysroot})
+    def update_ntdll(self):
```
ylatuya: missing line break before the fucntion

PR #158, opened 2018-09-27
Also check for winpthread library

nachogarglez:
In new versions of mingw clock_gettime and other time functions
are not longer part of the standard library, they have been
moved to wipthread library instead
This patch applies a patch into gstreamer's configure.ac in
order to check if clock_gettime is being provided by winpthread

### DIFF
--- a/recipes/gstreamer.recipe
+++ b/recipes/gstreamer.recipe
@@ -65,6 +65,7 @@ class Recipe(recipe.Recipe):
         if self.config.variants.nodebug:
             self.configure_options += ' --disable-gst-debug'
 
+	
     def extract(self):
         self.stype.extract(self)
         git.checkout(self.build_dir, 'HEAD')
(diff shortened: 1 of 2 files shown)


recipes/gstreamer.recipe
```diff
@@ -64,7 +64,7 @@ class Recipe(recipe.Recipe):
 
         if self.config.variants.nodebug:
             self.configure_options += ' --disable-gst-debug'
-
+
```
dabrain34: unnecessary tab

recipes/gstreamer.recipe
```diff
@@ -73,4 +73,7 @@ class Recipe(recipe.Recipe):
                 self.config.distro_version in [DistroVersion.UBUNTU_TRUSTY, DistroVersion.UBUNTU_UTOPIC, DistroVersion.UBUNTU_VIVID, DistroVersion.UBUNTU_WILY, DistroVersion.UBUNTU_XENIAL, DistroVersion.DEBIAN_STRETCH]):
             _patch = self.relative_path('gstreamer/0001-gst-parse-switch-to-lex-param.patch')
             shell.apply_patch(_patch, self.build_dir, 1)
+            _patch = self.relative_path('gstreamer/0002-also-check-for-winpthread.patch')
```
dabrain34: we should use the patches structure instead.

nachogarglez: I tried but for some reason that I didn't follow the repo is hard reset to HEAD *after* the patches are applied during the prepare stage.
I suspect that this is the reason of the former 0001 patch being there and I simply added a new patch below. It works that way....

nachogarglez: I'm holding this until the pending parts of INF-264 are ready to review

PR #170, opened 2018-11-27
Use the last revision of fluendo's gst-plugins-good

dabrain34:
Use fluendo's repo


dabrain34:
Not building on macOS with fluendo-devel.cbc:
xported_symbols_list,.libs/libgstisomp4-symbols.expsym
Undefined symbols for architecture x86_64:
  "_gst_codec_utils_aac_get_index_from_sample_rate", referenced from:
      _qtdemux_parse_trak in libgstisomp4_la-qtdemux.o
ld: symbol(s) not found for architecture x86_64

PR #176, opened 2019-01-30
Inf 330

dabrain34:


dabrain34:
Need additional work for the reverse dependencies...

dabrain34:
@ylatuya Could you please help us on how to manage this PR? Indeed in this commit, i was attempting to fulfill a missing check of dependencies. Indeed if i'm building fluendo-sdk which depends on libfludownloader, i gonna call "build fluendo-sdk" which will for the first time build fluendo-sdk and libfludownloader. But then the next day, if a change has been landed on libfludownloader, i'd like it to be rebuilt when i'm calling **build fluendo-sdk** and all the projects who are depending on libfludownloader.
Right now i used a 'build' argument (which can be renamed if its not clear) to keep the previous behaviour as default but use this parameter to see what are the advantages and the drawbacks of this new behaviour and enhance if necessary.
By example right now the reverse dependencies is not working as expected but i need time to debug it and fix it which I dont have :)

dabrain34:
new PR available https://github.com/fluendo/cerbero/pull/177

PR #177, opened 2019-02-20
Add fetch argument to build command.

dabrain34:
it allows to check all recipes source state and invalidate the build
if necessary on each of those where the source state has changed.

### DIFF
--- a/cerbero/commands/build.py
+++ b/cerbero/commands/build.py
@@ -18,7 +18,7 @@
 
 
 #from cerbero.oven import Oven
-from cerbero.commands import Command, register_command
+from cerbero.commands import Command, register_command, fetch
 from cerbero.packages.packagesstore import PackagesStore
 from cerbero.build.oven import Oven
 from cerbero.utils import _, N_, ArgparseArgument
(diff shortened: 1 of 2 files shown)


dabrain34:
To replace https://github.com/fluendo/cerbero/pull/176

cerbero/commands/build.py
```diff
@@ -92,9 +94,14 @@ def runargs(self, config, recipes, missing_files=False, force=False,
             ordered_recipes = recipes
         else:
             ordered_recipes = cookbook.list_recipes_deps(recipes)
+        if fetch:
+          self._fetch(config, ordered_recipes, no_deps=False, reset_rdeps=True)
         oven.start_cooking(ordered_recipes, use_binaries, upload_binaries,
                 build_missing)
 
+    def _fetch(self, config, recipes, no_deps=False, reset_rdeps=False):
```
ylatuya: no_deps and reset_rdeps are not needed as it's always used with the same values.

dabrain34: Yes I prefered it moludar

dabrain34: > We should try to upstream this first

Ok

ylatuya: yes, we are already upstreaming things, I don't understand the question :) GitLab's account and fluendo's email in the commits.

PR #178, opened 2019-03-18
Add package variables for wix installer

dfernandezramos:


cerbero/packages/wix.py
```diff
@@ -551,7 +553,9 @@ def _customize_ui(self):
                 etree.SubElement(self.product, 'WixVariable',
                         Id='WixUI%s' % var, Value=path)
         # Icon
-        path = self.package.relative_path(self.ICON)
+        path = self.package.package_subdir + self.ICON
```
vguzmanp: use `os.path.join`

cerbero/packages/wix.py
```diff
@@ -380,8 +380,10 @@ def _init_replacements(self):
             "@Platform@": self._platform(),
             "@UIType@": self.ui_type,
             "@CerberoPackageDir@": self.package.package_dir(),
+            "@CerberoPackageSubDir@": self.package.package_subdir,
             "@CerberoPrefixDir@": self.prefix,
             "@CustomVariables@": self._custom_variables(),
+            "@ExecutableName@": self.package.executable_name,
```
vguzmanp: Here you should get `self.package.executable_name [Platform.Windows]`

cerbero/packages/package.py
```diff
@@ -94,10 +94,16 @@ class PackageBase(object):
...
     @type rpm_provides = str
+    @cvar package_subdir = package subdirectory for beta resources
+    @type package_subdir = str
+    @cvar executable_name = package exectutable name used in wix installer.wxs
+    @type executable_name = str
     '''
     name = 'default'
     shortdesc = 'default'
     longdesc = 'default'
+    package_subdir = None
```
vguzmanp: Initialize this to `''` so that it doesn't crash os.path.join

PR #180, opened 2019-03-26
RA-2264

dfernandezramos:


packages/vas-deps.package
```diff
@@ -9,11 +9,8 @@ class Package(package.Package):
     longdesc  = 'Video Analysis Solution dependencies'
     files = ['ubuntu-font-family:fonts',
              'raleway-font-family:fonts',
-             'pango:fonts',
-             'gtk-sharp',
              'libglade:libs',
              'sqlite3:libs',
-             'fontconfig:bins',
```
vguzmanp: Now that we don't depend on fontconfig, we should remove the fc-cache call from the installer and from the splash screen 🎉

vguzmanp: libglade is from gtk too

PR #182, opened 2019-04-09
keep the original PS1 info from host env

dabrain34:


cerbero/utils/shell.py
```diff
@@ -548,7 +548,7 @@ def enter_build_environment(platform, arch, sourcedir=None):
 source ~/.bashrc
 fi
 %s
-PS1='\[\033[01;32m\][cerbero-%s-%s]\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
```
ylatuya: Isn't new format loosing the ':' that splits the cerbero hint and the actual PS1 ?

PR #190, opened 2019-06-06
wix: add support for msi creation in cross-compilation

ylatuya:
Make use of fragments instead of merge module which
can't be created due to a bug in wine

### DIFF
--- a/config/windows.config
+++ b/config/windows.config
@@ -21,6 +21,7 @@ if platform == Platform.WINDOWS:
     if target_arch == Architecture.X86_64:
         build = 'x86_64-w64-mingw32'
 else:
+    os.environ['WIX'] = os.path.join(build_tools_prefix, 'lib', 'wix')
     separator = ':'
 
 if target_arch == Architecture.X86:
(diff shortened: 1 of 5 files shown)


config/windows.config
```diff
@@ -21,6 +21,7 @@ if platform == Platform.WINDOWS:
     if target_arch == Architecture.X86_64:
         build = 'x86_64-w64-mingw32'
 else:
+    os.environ['WIX'] = os.path.join(prefix, 'lib', 'wix')
```
neodesys: Just to get sure: prefix here points to build-tools ? or is it the cerbero prefix (to dist/arch directory) ?

neodesys: I've tried it and it is from build-tools. Setting this:
os.environ['WIX'] = os.path.join(prefix, '..', '..', 'build-tools', 'lib', 'wix')
works, but if you have a variable pointing directly to build-tools folder it would be better than my patch ;)

dabrain34:
I was still having some issues with the wine install with gitlab's cerbero, do you think you could upstream this ?

PR #191, opened 2019-06-13
Wix fragments

neodesys:


dabrain34:
Hello, do you think you gonna upstream this branch later ? In this case I can drop the custom branch I was workin on during the Hackfest

neodesys:
Yes, I will create a MR on upstream as soon as this PR is reviewed and accepted on our side :+1:

cerbero/packages/wix_packager.py
```diff
@@ -57,17 +57,17 @@ def pack(self, output_dir, devel=False, force=False, keep_temp=False):
         return paths
 
     def create_merge_module(self, output_dir, package_type, force, version,
-                            keep_temp):
+                            keep_temp, keep_strip_temp_dir=False):
```
ylatuya: I don't think you need this variable, you only have to keep it when using fragments

cerbero/packages/wix.py
```diff
@@ -222,28 +301,28 @@ def _add_root_dir(self):
 
     def _add_vs_templates(self):
         etree.SubElement(self.module, 'PropertyRef',
-            Id='VS_PROJECTTEMPLATES_DIR')
+                         Id='VS_PROJECTTEMPLATES_DIR')
```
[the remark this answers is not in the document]
neodesys: They were not, it's pep8 formatting automatically when executed on save.
If it's an issue, I can restablish previous indentation ?

cerbero/packages/wix_packager.py
neodesys: Yes, but as the method returns different kind of values depending on this variable value (a single string or a duple of strings), I prefer keeping it explicit on method parameters than depending on an internal state.

So you know exactly what kind of data to expect as a return value at the moment of calling the method.

PR #195, opened 2019-10-01
Fix Windows to POSIX path conversion

mjimenofluendo:


cerbero/config.py
```diff
@@ -279,7 +279,7 @@ def get_env(self, prefix, libdir, py_prefix):
         manpathdir = os.path.join(prefix, 'share', 'man')
         infopathdir = os.path.join(prefix, 'share', 'info')
         pkgconfigbin = os.path.join(self.build_tools_prefix, 'bin', 'pkg-config')
-        pkgconfigdatadir = os.path.join(prefix, 'share', 'pkgconfig')
+        pkgconfigdatadir = to_unixpath(os.path.join(prefix, 'share', 'pkgconfig'))
```
neodesys: Cerbero is already managing / and \ internally with path (it has patched the path character).
It is really strange that all other paths work except this one.
Are you sure that is not some mess up between native mingw build and wsl ?
Are you sure this change is not breaking the build on linux-native, linux-to-windows-cross-compilation ?

PR #200, opened 2019-11-08
Create mingw/multilib/x86_64-w64-mingw32/include on bootstrap

mjimenofluendo:
The mingw-6.0.0-gcc-8.2.0-windows-multilib.tar.xz files didn't contain an include folder and we change the Windows.py for creating it

### DIFF
--- a/cerbero/bootstrap/windows.py
+++ b/cerbero/bootstrap/windows.py
@@ -131,6 +131,12 @@ class WindowsBootstrapper(BootstrapperBase):
             sysroot = os.path.join(self.prefix, 'x86_64-w64-mingw32/sysroot')
             shutil.rmtree(os.path.join(sysroot, 'mingw'))
             shell.call('ln -s usr/x86_64-w64-mingw32 mingw', sysroot)
+            # Create include folder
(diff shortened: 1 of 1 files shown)


cerbero/bootstrap/windows.py
```diff
@@ -131,6 +131,9 @@ def fix_mingw(self):
             sysroot = os.path.join(self.prefix, 'x86_64-w64-mingw32/sysroot')
             shutil.rmtree(os.path.join(sysroot, 'mingw'))
             shell.call('ln -s usr/x86_64-w64-mingw32 mingw', sysroot)
+            # Make include folder
+            include_folder = os.path.join(self.prefix, 'x86_64-w64-mingw32/include')
+            os.makedirs( include_folder )
```
neodesys: It compiles with an empty include folder ?
os.makedirs() doesn't complain about the include file created when unpacking the tar.gz file ?
In Cerbero there is no spaces after/before parentheses.

mjimenofluendo: In one step the include folder is a 30 kb file, how now we have a folder the folder there isn't empty, though that it works with an empty folder.
The .tar file doesn't contain an include folder, but I include a try..except

PR #212, opened 2019-12-03
Sign dll and packages by overwriting the sign_dll function of the Pac…

mjimenofluendo:
…kage class

### DIFF
--- a/cerbero/packages/wix_packager.py
+++ b/cerbero/packages/wix_packager.py
@@ -26,7 +26,7 @@ from cerbero.packages import PackagerBase, PackageType
 from cerbero.packages.package import Package, App
 from cerbero.utils import messages as m
 from cerbero.utils import shell, to_winepath, get_wix_prefix
-from cerbero.tools import strip
+from cerbero.tools import strip, sign
 from cerbero.packages.wix import MergeModule, VSMergeModule, MSI, WixConfig, Fragment
 from cerbero.packages.wix import VSTemplatePackage
 from cerbero.config import Platform
@@ -67,7 +67,7 @@ class MergeModulePackager(PackagerBase):
         tmpdir = None
         # For packages that requires stripping object files, we need
         # to copy all the files to a new tree and strip them there:
-        if self.package.strip:
+        if self.package.strip or self.package.wix_sign_dll:
             tmpdir = tempfile.mkdtemp()
             for f in files_list:
                 src = os.path.join(self.config.prefix, f)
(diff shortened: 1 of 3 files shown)


cerbero/packages/wix_packager.py
```diff
@@ -27,6 +27,7 @@
 from cerbero.utils import messages as m
 from cerbero.utils import shell, to_winepath, get_wix_prefix
 from cerbero.tools import strip
+from cerbero.tools import sign
```
pamarcos: You can import this on the same line as the previous one:
`from cerbero.tools import strip, sign`

cerbero/tools/sign.py
```diff
@@ -0,0 +1,45 @@
...
+        for f in self.excludes:
+            if f in path:
+                return
+        try:
+            self.package.sign_dll(path)
+        except:
+            pass
+
+    def sign_dir(self, dir_path):
+        for dirpath, dirnames, filenames in os.walk(dir_path):
```
pamarcos: Here you are not using `dirnames`, I think you can use `_` to explicitly tell that you are not interested in that variable.
However, `sign_dir` to me makes me think that it will recursively sign every file in every subdirectory as well. So, maybe `dirnames` should actually be used as well to sign each of them recursively?

cerbero/tools/sign.py
```diff
@@ -0,0 +1,45 @@
...
+import os
+from cerbero.config import Platform
+from cerbero.utils import shell
+
+
+class Sign(object):
+    '''Wrapper for the sign tool'''
+
+    def __init__(self, config, package):
+        self.config = config
```
pamarcos: `config` doesn't seem to be used at all here

cerbero/tools/sign.py
```diff
@@ -0,0 +1,45 @@
...
+        self.excludes = package.strip_excludes
+        self.package = package
+
+    def sign_file(self, path):
+        for f in self.excludes:
```
pamarcos: Do you need to be greedy here? Meaning being `asd/esd` the file `f` and `this/is/asd/esd/isd/test.txt` the `path`, this will consider the file is excluded. In case we do not need to be greedy, we could simply use `if path in self.excludes` that will match only if any of the entries match exactly what's in `self.excludes`.

cerbero/tools/sign.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Library General Public License for more details.
+#
+# You should have received a copy of the GNU Library General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
```
pamarcos: There's no need to include the shebang for python3. Also, the copyright should be to Fluendo, being you the author. Something like this:
```
# cerbero - a multi-platform build system for Open Source software
# Copyright (C) 2019, Fluendo, S.A.
#  Author: Pablo Marcos Oltra <pmarcos@fluendo.com>, Fluendo, S.A.
#
# This library is free software; you can redistribute it and/or
# modify it under the terms of the GNU Library General Public
# License as published by the Free Software Foundation; either
# version 2 of the License, or (at your option) any later version.
#
# This library is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
# Library General Public License for more details.
#
# You should have received a copy of the GNU Library General Public
# License along with this library; if not, write to the
# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
# Boston, MA 02111-1307, USA.
```

cerbero/packages/wix_packager.py
mjimenofluendo: Done

cerbero/tools/sign.py
mjimenofluendo: It is not used anywhere, I have removed it

cerbero/tools/sign.py
mjimenofluendo: I changed the text

PR #218, opened 2019-12-18
Fridge 1.0

pamarcos:
# CI

http://172.16.10.104:8080/job/ci_test/ - Linux, macOS and cross compilation from Linux for Windows
http://172.16.10.104:8080/job/Fridge_ci_test/ - Native Windows

# Fridge

The fridge functionality of Cerbero aims to reduce the time taken to cook
recipes by freezing them once they are built and uploading their content to some
FTP server that acts as a fridge for the recipes. This way, we avoid the
configure and compile steps, which are arguably the most time-consuming ones.

## How recipes are reused

A hash is generated for the current environment used in the configuration. This
means everything that `config::get_env` considers plus the compiler version.
e.g. the linker flags, include dir, LD_LIBRARY_PATH, etc. Hence, the idea is to
reuse the same binaries already built in a conservative way to ensure they will
work. The environment is sanitized before creating the hash, so that the prefix,
Cerbero's home and user's home are normalized. Apart from that, the package
naming already contains the target platform and target architecture. A directory
is created for each of the configurations (aka. environments) where prebuilt
packages will live.

Apart from the environment hash, a hash is generated per recipe to allow having
different versions. The recipe hash is taken from a checksum of all the files
involving it (the recipe itself + patches) and the version that it's using. For
recipes using a git repo as a source that means the commit it's based upon.

# What is packaged

Initially, we had a *devel* and *release* version of each package and fridge was
using the same files that are specified in the recipes that need to be packaged.
It turned out to be very troublesome and difficult to maintain a list of files
needed for each recipe that is used during the bootstrap step for each OS. As a
consequence, a pragmatic approach was taken in which we only have *devel*
packages for each recipe which includes every single file that is actually
installed. This way, we ensure that what's packaged has everything needed. The
source code is prepared to switch to the *devel* + *release* flavours at any
point, though.

The installed files collection is done by gathering all the files that *install*
and *post_install* add or modify. This list of files is added to the cache into
a new `installed_files` member. This may be used in the future not only to know
which files fridge needs to package, but also to fully uninstall all the files a
recipe has added.

## How recipes are stored in the fridge

The fridge consists of a FTP server with write permissions granted so that the
client can create a different directory per configuration. This directory will
contain all the packages built for this specific configuration for all target
platforms and target architectures.

Every frozen recipe is uploaded along a sha256 files containing the hash of the
packages file. This file will be used on the local fridge to know whether the
remote one needs to be downloaded or not. In other words, the `binaries_local`
directory specified in the config acts as a local cache for the frozen recipes.

Additionally, every environment directory includes an `ENVIRONMENT` file
containing the list of all the variables considered to generate the hash:

```
9beba1f21a679d67ce08ce2a745d0492

ACLOCAL=aclocal
ACLOCAL_FLAGS=-I{PREFIX}/share/aclocal
CC=clang
CC_VERSION=10.0.0
CFLAGS=-Wall -g -O2 -arch x86_64 -m64 -Wno-error=format-nonliteral -I{PREFIX}/include -mmacosx-version-min=10.10 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk
CPLUS_INCLUDE_PATH={PREFIX}/include
C_INCLUDE_PATH={PREFIX}/include
GI_TYPELIB_PATH={PREFIX}/lib/girepository-1.0
GSTREAMER_ROOT={PREFIX}
GST_PLUGIN_PATH={PREFIX}/lib/gstreamer-0.10
GST_PLUGIN_PATH_1_0={PREFIX}/lib/gstreamer-1.0
GST_REGISTRY={USER}/.gstreamer-0.10/cerbero-registry-x86_64
GST_REGISTRY_1_0={USER}/.cache/gstreamer-1.0/cerbero-registry-x86_64
INFOPATH={PREFIX}/share/info
LDFLAGS=-L{PREFIX}/lib -headerpad_max_install_names -Wl,-headerpad_max_install_names -Wno-error=unused-command-line-argument -arch x86_64 -m64 -Wl,-arch,x86_64 -mmacosx-version-min=10.10 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk
LD_LIBRARY_PATH={PREFIX}/lib:{PREFIX}/lib:{PREFIX}/bin:/opt/local/bin:/opt/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/share/dotnet:~/.dotnet/tools:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Applications/Xamarin Workbooks.app/Contents/SharedSupport/path-bin
MANPATH={PREFIX}/share/man
MONO_GAC_PREFIX={PREFIX}
MONO_PATH={PREFIX}/lib/mono/4.5
PERL5LIB={PREFIX}/lib/perl5:{PREFIX}/lib/perl5/site_perl/5.18.2
PKG_CONFIG={PREFIX}/bin/pkg-config
PKG_CONFIG_LIBDIR={PREFIX}/lib/pkgconfig
PKG_CONFIG_PATH={PREFIX}/share/pkgconfig
PYTHONPATH={PREFIX}/lib/python3.7/site-packages/:{PREFIX}/lib/python3.7/site-packages/
XCURSOR_PATH={PREFIX}/share/icons
XDG_CONFIG_DIRS={PREFIX}/etc/xdg
XDG_DATA_DIRS={PREFIX}/share
```

The following directory contains the packages of yasm and zlib recipes for macOS
64bit, along with their sha256:

```
└── ffdfa03d41ab86a88f559ab46546c7ac
    ├── ENVIRONMENT
    ├── yasm-pkg-darwin-x86_64-1.3.0-aa98cab175d302577087dcaa225f4df3-devel.tar.bz2
    ├── yasm-pkg-darwin-x86_64-1.3.0-aa98cab175d302577087dcaa225f4df3-devel.tar.bz2.sha256
    ├── zlib-pkg-darwin-x86_64-1.2.11-dffcadaf533a8451920690663494d892-devel.tar.bz2
    ├── zlib-pkg-darwin-x86_64-1.2.11-dffcadaf533a8451920690663494d892-devel.tar.bz2.sha256
```

## How to use the fridge functionality

First, we need add the `binaries_remote` element into the configuration. e.g.:
`binaries_remote = FtpBinaryRemote('127.0.0.1:2121')`.

Apart from that, the following arguments have been added to the `build` and
`bootstrap` commands:

```
--use-binaries     use binaries from the repo before building
--upload-binaries  after a recipe is built upload the corresponding binary package
--build-missing    in case a binary package is missing try to build it
--fridge           equivalent to '--build-missing --use-binaries --upload-binaries'
```

As a rule of thumb, if you want to use prebuilt binaries and upload any of them
that are not already present, you would use `--fridge`. Cerbero will
automatically use the ones that exist but build the ones who doesn't, uploading
them to the fridge if the option is given.


cerbero/build/filesprovider.py
```diff
@@ -123,8 +123,8 @@ class FilesProvider(object):
     # UNIX shared libraries can have between 0 and 3 version components:
     # major, minor, micro. We don't use {m,n} here because we want to capture
     # all the matches.
-    _SO_REGEX = r'^lib{}\.so(\.[0-9]+)?(\.[0-9]+)?(\.[0-9]+)?$'
-    _DYLIB_REGEX = r'^lib{}(\.[0-9]+)?(\.[0-9]+)?(\.[0-9]+)?\.dylib$'
+    _SO_REGEX = r'^{}\.so((\.[0-9]+)+)?$'
```
ylatuya: Is this change because of dll's not starting with lib? There is a specific regex expresion for DLL's IIRC

cerbero/build/relocatabletar.py
```diff
@@ -0,0 +1,59 @@
...
+# with relocation abilities
+#------------------------------------------------------------------------------
+
+import os
+from tarfile import TarFile
+from cerbero.utils import shell, is_text_file
+from cerbero.tools.osxrelocator import OSXRelocator
+
+class RelocatableTar(TarFile):
+#------------------------------------------------------------------------------
```
ylatuya: Use correct python docs

cerbero/packages/disttarball.py
```diff
@@ -121,41 +141,101 @@ def _create_tarball_stripped(self, output_dir, package_type, files, force,
...
+                        fo.write(content)
+                        fo.truncate()
+                        fo.flush()
+                        tar_files.append(filepath)
+                else:
+                    tar_files.append(filepath)
+            else:
+                tar_files.append(filepath)
+
+        # This link allows to rpm packager to install the libs
```
ylatuya: Is this really needed? For RPM target distros lib_suffix = 64 should be used in the .cbc
Otherwise a lib32 link should also be needed

cerbero/config.py
```diff
@@ -511,6 +519,44 @@ def target_distro_version_gte(self, distro_version):
         assert distro_version.startswith(self.target_distro + "_")
         return self.target_distro_version >= distro_version
 
+    @lru_cache(maxsize=None)
+    def _get_sanitized_env(self):
```
ylatuya: private functions should be ordered after public ones

cerbero/build/cookbook.py
```diff
@@ -55,25 +55,30 @@ class RecipeStatus (object):
...
 
     def touch(self):
         ''' Touches the recipe updating its modification time '''
         self.mtime = time.time()
 
     def __repr__(self):
-        return "steps: %r, needs_build: %r, mtime: %r, filepath: %r, built_version: %r, file_hash: %r" % \
-            (self.steps, self.needs_build, self.mtime, self.filepath, self.built_version, self.file_hash.hex())
+        return "steps: %r, needs_build: %r, mtime: %r, filepath: %r, built_version: %r, file_hash: %r\ninstalled_files: %r" % \
+            (self.steps, self.needs_build, self.mtime, self.filepath, self.built_version, self.file_hash,
```
neodesys: I don't know what it is changing but you removed the .hex() from self.file_hash

cerbero/build/fridge.py
```diff
@@ -0,0 +1,279 @@
...
+            # in which case, we won't generate it again unless force is used
+            if step in [self.GEN_BINARY[1], self.UPLOAD_BINARY[1]]:
+                if self.cookbook.step_done(recipe.name, self.EXTRACT_BINARY[1]) and not self.force:
+                    m.action(_('No need since a package exists in remote for this recipe'))
+                    self.cookbook.update_step_status(recipe.name, step)
+                    continue
+
+            # call step function
+            stepfunc = getattr(self, step)
+            if not stepfunc:
```
neodesys: If it is possible that self has not attribute for step, you should use hasattr() before getattr() because if there is no self[step] getattr() will throw an AttributeError().

In this code your are checking that self[step] can be None or a function but you implicitly consider that step is always defined as an attribute of self.

Anyway, in all cases you will end up with a raised error, so result is the same.

cerbero/build/filesprovider.py
pamarcos: This comes from commit 237c5404a13628871e85effac027bcb101fd286b:
>There are some library names that do not start with "lib", such as preloadable_libintl.so

Filesprovider is irrelevant now for fridge since we don't use it as we package every single file that is installed, but the regex shouldn't be as strict because there are quite a few exceptions.

cerbero/build/oven.py
```diff
@@ -141,11 +151,29 @@ def _cook_recipe(self, recipe, count, total):
         if not self.cookbook.recipe_needs_build(recipe.name) and \
                 not self.force:
             m.build_step(count, total, recipe.name, _("already built"))
+
+            if upload_binaries:
+                try:
+                    fridge.freeze_recipe(recipe, count, total)
+                except RecipeNotFreezableError:
+                    pass
```
neodesys: Maybe we could log an INFO when a recipe is not freezable.

cerbero/build/relocatabletar.py
pamarcos: Sure. This file comes directly from master: https://github.com/fluendo/cerbero/blob/master/cerbero/build/relocatabletar.py
I did a small modification, though

cerbero/build/oven.py
```diff
@@ -189,12 +217,32 @@ def _cook_recipe(self, recipe, count, total):
...
 
         if self.missing_files:
             self._print_missing_files(recipe, tmp)
-            tmp.close()
+
+        if upload_binaries:
+            try:
+                fridge.freeze_recipe(recipe, count, total)
+            except RecipeNotFreezableError:
+                pass
```
neodesys: Same comment about logging an INFO if not freezable

cerbero/packages/disttarball.py
pamarcos: I have no idea, to be honest. That specific chunk is taken directly from https://github.com/fluendo/cerbero/blame/master/cerbero/packages/distarchive.py#L142-L153 and was added by Stephane here https://github.com/fluendo/cerbero/commit/d25a81422411192309a8def5b363e586a0e43fc8

cerbero/commands/fetch.py
```diff
@@ -58,14 +61,23 @@ def fetch(cookbook, recipes, no_deps, reset_rdeps, full_reset, print_only):
...
                 if isinstance(recipe, Tarball):
                     m.message("TARBALL: {} {}".format(recipe.url, recipe.tarball_name))
                 continue
-            m.build_step(i + 1, len(fetch_recipes), recipe, 'Fetch')
+            if fridge:
+                try:
+                    fridge.fetch_recipe(recipe, i + 1, len(fetch_recipes))
+                    continue
+                except Exception:
+                    pass
```
neodesys: Maybe some logging ?

cerbero/build/cookbook.py
pamarcos: Yeah, I changed the type of `file_hash` from int to str on purpose

PR #227, opened 2020-01-27
Lazy load packages from recipes for fridge

pamarcos:
Since these packages should only be used for fridge anyways, we'd rather lazy load them only when needed to avoid the extra work at PackageStore initialization.


pamarcos:
> With this patch virtual-packages (recipe + `-pkg`) are not listed with the `list-packages` command.

Yes, this is what you expected, right? The reason they're not listed is basically because they have not even been parsed (yet). They will whenever they are needed for Fridge, though.

PR #230, opened 2020-02-04
Ope 499

mjimenofluendo:


cerbero/packages/linux.py
```diff
@@ -54,9 +54,12 @@ def pack(self, output_dir, devel=True, force=False, keep_temp=False,
         if isinstance(self.package, App) and self.package.embed_deps:
             pass
         elif pack_deps:
-            self.pack_deps(output_dir, tmpdir, force)
+            self.pack_deps(output_dir, tmpdir, force, split)
 
-        if not isinstance(self.package, MetaPackage):
+        if isinstance(self.package, MetaPackage) and not self.package.build_meta_package:
+            return []
```
pamarcos: What's the purpose of returning an empty list instead of simply returning? It may be worth adding a log trace here to know why the pack did nothing, unless is a very common scenario

cerbero/packages/rpm.py
```diff
@@ -164,23 +164,26 @@ def setup_source(self, tarball, tmpdir, packagedir, srcdir):
...
+    def prepare(self, tarname, tmpdir, packagedir, srcdir, split):
         try:
-            runtime_files = self._files_list(PackageType.RUNTIME)
+            runtime_files = self._files_list(PackageType.RUNTIME, split)
         except EmptyPackageError:
             runtime_files = ''
 
-        if runtime_files or isinstance(self.package, MetaPackage):
+        if self.install_dir not in ['/usr','/usr/local']:
+            runtime_files = runtime_files + '\n'+os.path.join(self.install_dir,'')
```
pamarcos: whitespaces missing surrounding `+`

cerbero/packages/linux.py
mjimenofluendo: The fluendo-sdk-installer, without --no-split option,  it's creating three packages, the binary package, the development files package, and a package that contains dependencies for both, when we specified the --no-split option, we don't want to create the meta package, all is in one file, that is why we use the empty list return to disable the creation of the meta package.

cerbero/packages/rpm.py
mjimenofluendo: fixed up

PR #239, opened 2020-02-20
cmake: Fix error with cmake recipes with use_system_libs

rgonzalezfluendo:
Error:
```
Traceback (most recent call last):
  File "./cerbero/cerbero/build/oven.py", line 202, in _cook_recipe
    loop.run_until_complete(stepfunc(recipe))
  File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
    return future.result()
  File "./cerbero/cerbero/build/recipe.py", line 54, in wrapped
    await ret
  File "./cerbero/cerbero/build/build.py", line 170, in call
    res = await func(*args)
  File "./cerbero/cerbero/build/build.py", line 573, in configure
    await MakefilesBase.configure(self)
  File "./cerbero/cerbero/build/build.py", line 168, in call
    self._add_system_libs()
  File "./cerbero/cerbero/build/build.py", line 327, in _add_system_libs
    self.set_env(var, val)
  File "./cerbero/cerbero/build/build.py", line 260, in set_env
    self.check_reentrancy()
  File "./cerbero/cerbero/build/build.py", line 247, in check_reentrancy
    raise RuntimeError('Do not modify the env inside @modify_environment, it will have no effect')
RuntimeError: Do not modify the env inside @modify_environment, it will have no effect
```

### DIFF
--- a/cerbero/build/build.py
+++ b/cerbero/build/build.py
@@ -531,7 +531,6 @@ class CMake (MakefilesBase):
                     '-DCMAKE_FIND_ROOT_PATH=$CERBERO_PREFIX '\
                     '-DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=true '
 
-    @async_modify_environment
     async def configure(self):
         cc = self.env.get('CC', 'gcc')
         cxx = self.env.get('CXX', 'g++')


pamarcos:
I need more context here since this decorator is new to me and I don't know all the implications of removing it from the `configure` step and how it may affect the rest. Probably @ylatuya knows better.

rgonzalezfluendo:
The probles is that `@async_modify_environment` is called twice:
 * In `CMake.configure` (which I deleted in this PR)
 * In `MakefilesBase.configure`. Directly called at the end of `CMake.configure`: https://github.com/fluendo/cerbero/blob/5219111228beaf441698e76a27dc9cc732707225/cerbero/build/build.py#L572

pamarcos:
> The problem is that `@async_modify_environment` is called twice:
> 
>     * In `CMake.configure` (which I deleted in this PR)
> 
>     * In `MakefilesBase.configure`. Directly called at the end of `CMake.configure`: https://github.com/fluendo/cerbero/blob/5219111228beaf441698e76a27dc9cc732707225/cerbero/build/build.py#L572

i see. I'd have expected to have this `modify_environment` thingy at the top. In this case, in `CMake.configure` but then not in `MakefilesBase.configure`. For what I see in the code, the `CMake.configure` is not making any cmd call, so we should be fine. But the moment someone calls something within the `CMake.configure`, we won't be using the proper environment.

Ideally, I'd like to have a generic solution that ensures that the env is set correctly for all cases, but I also lack the context for this to make it properly

marecki315:
I've created a proper fix in https://github.com/fluendo/cerbero/pull/243.
I recommend close this PR.

PR #240, opened 2020-02-20
Use builddir when building CMake recipes

marecki315:


pamarcos:
I was in the daily when you requested my review. The code seems fine, but maybe the `make_dir` name is a little confusing to me. I'd have used `cmake_builddir` following the `meson_builddir` approach.

pamarcos:
> I was in the daily when you requested my review. The code seems fine, but maybe the `make_dir` name is a little confusing to me. I'd have used `cmake_builddir` following the `meson_builddir` approach.

Nevermind. I see you reuse the `make_dir` in `MakefilesBase`.

PR #247, opened 2020-02-26
Ope 677

neodesys:


cerbero/tools/depstracker.py
```diff
@@ -85,7 +104,8 @@ def __init__(self, platform, prefix):
         self.lister = self.BACKENDS[platform]()
 
     def list_deps(self, path):
-        deps = self.lister.list_deps(self.prefix, path)
+        loop = asyncio.get_event_loop()
+        deps = loop.run_until_complete(self.lister.async_list_deps(self.prefix, path))
```
ylatuya: What is the point on converting these functions to async if they are not being parallelized at any point?

cerbero/tools/strip.py
```diff
@@ -18,33 +18,53 @@
...
     def strip_dir(self, dir_path):
+        if not self.strip_cmd:
+            m.warning('Strip command is not defined')
+            return
+
+        tasks = []
         for dirpath, dirnames, filenames in os.walk(dir_path):
             for f in filenames:
-                self.strip_file(os.path.join(dirpath, f))
+                tasks.append(self.async_strip_file(os.path.join(dirpath, f)))
```
ylatuya: This is a bit dangerous, you are parallelizing without limiting the numbers of parallel tasks. If you have 4000 files you will run 4000 process all together.

cerbero/packages/wix.py
```diff
@@ -111,8 +118,12 @@ def _format_path_id(self, path, replace_dots=False):
         return ret
 
     def _format_dir_id(self, string, path, replace_dots=False):
-        return self._format_id(string, replace_dots) + '_' +\
+        ret = self._format_id(string, replace_dots) + '_' +\
             self._format_path_id(path, replace_dots)
+        # Wix Id length is limited to 72 characters.
+        if len(ret) > 72:
```
ylatuya: Isn't this the same trunking algorithm used in _format_path_id ?

cerbero/packages/linux.py
```diff
@@ -104,21 +105,24 @@ def build(self, output_dir, tarname, tmpdir, packagedir, srcdir):
...
+            dep_paths = []
             try:
-                paths = packager.pack(self.output_dir, self.devel, self.force, True, self.split, True, tmpdir)
+                dep_paths = packager.pack(self.output_dir, self.devel, self.force, True, self.split, True, tmpdir)
             except EmptyPackageError:
                 self._empty_packages.append(p)
-                paths = []
-            p.post_package(paths, self.output_dir)
+            finally:
+                paths += p.post_package(dep_paths, self.output_dir)
```
ylatuya: post_package does not return a new list of packages to be appended, post_package takes the current list of packages, modifies it and return the new list of packages

cerbero/tools/depstracker.py
neodesys: Mainly to get rid of this warning:
Exception ignored when trying to write to the signal wakeup fd:
BlockingIOError: [Errno 11] Resource temporarily unavailable
which needs python >= 3.8 to be corrected

cerbero/packages/wix.py
neodesys: not exactly because _format_path_id() may add a number after, so I reserve 3 more characters.

cerbero/tools/strip.py
neodesys: This is exactly what shell.run_until_complete(tasks) does. It limits by default to 16 `max_concurrent=16` concurrent tasks.

cerbero/packages/linux.py
neodesys: post_package takes dep_paths as argument, the paths variable is accumulating files for all deps.

cerbero/packages/wix.py
ylatuya: format_path_id adds a number to avoid an ID collision having 2 ids that are identical. When trunking the ID to X characters if the X characters are identical you will have 2 id's that are similar.
The logic that trunks the ID should be similar in all places, where you trunk the ID to have X character and use a number to avoid duplicate ID's.

Take this as example with a max length of 10:
12345678910ABC
12345678910XYZ

If you trunk this at 10 you end up with a collision:
12345678910
12345678910

The  algorithm should instead do:
1234567_00
1234567_01

cerbero/tools/depstracker.py
ylatuya: Just to understand, who is raising this error, shell.check_call ?
I get the interest of making the API async, but it's useless has if  DepsTracker.list_deps is not async itself, the DepsTracker will still be synchronous.
It's redundant to prefix functions with async_, this is normally only used when there are 2 variants of the function async and not async.

cerbero/packages/wix.py
neodesys: You're right, I'm gonna improve it in a fix OPE-677-Fixes branch.

cerbero/tools/depstracker.py
neodesys: Yes this error is raised by shell.check_call().
In fact all subprocess methods are raising it when called more than 300 times/sec (which is the case when doing a lot of strip or ldd while building a package). This warning is fixed in Python >= 3.8 but for now it pollutes a lot the logs and may hide other warnings more important.

In the upstream version of Cerbero the whole system goes towards asynchronous calls for system calls, so I begun to adapt it here to have it ready for future usage and get rid of the warning at the same time.

I didn't want to spend time modifying all the callers, so it's true that for now, having it asynchronous has very few interest from a performance point of view.

PR #254, opened 2020-03-14
Create a .versions file in the package command

ylatuya:
This files contains the list of recipes and the versions
used to build them. This helps debugging possible issues
during packaging


ylatuya:
I will rewrite this into a new command

PR #256, opened 2020-03-16
Use the correct pack args based on the package type

ylatuya:
This should never happen, all implementation should have the same
method signature. If someone needs a different pack function
it should create a new method


ylatuya:
I will discard this one in favor of  https://github.com/fluendo/cerbero/pull/255 which as not merged

PR #260, opened 2020-03-18
COM-1113 Clean old installed files and implement a proper uninstall command

pamarcos:


cerbero/build/cookbook.py
```diff
@@ -196,14 +196,27 @@ def update_installed_files(self, recipe_name, files):
         @param files: installed files
         @type files: list
         '''
+        # os.path.exists returns False for broken symbolic links
+        def _file_exists(install_dir, file):
+            f = os.path.join(install_dir, file)
+            return os.path.exists(f)
```
neodesys: why don't you use os.path.isfile() here ? because if there is a directory os.remove() below will fail anyway.

cerbero/utils/shell.py
```diff
@@ -782,3 +782,12 @@ def run_until_complete(tasks, max_concurrent=16):
...
     return result
+
+
+def remove_empty_dirs(path):
+    '''
+    Remove all empty dirs within path
+    @param path: the path to remove empty dirs from
+    @type path: str
+    '''
+    new_call('find {} -type d -delete'.format(path))
```
neodesys: If there are some non-empty dirs, it will output errors and `find` tool will return 1 as error code.
Maybe it is not an issue, I just wanted to state it out just in case ;)

pamarcos: Good catch! I'll add the `-empty` argument to sort it out

PR #262, opened 2020-03-24
COM-1107 Opt-in for considering parents for recipe's checksum

pamarcos:


cerbero/build/recipe.py
```diff
@@ -733,7 +720,8 @@ def get_checksum(self):
         @return: a checksum of the recipe file and its dependencies
         @rtype: str
         '''
-        sha256 = self._get_parents_checksum()
+        sha256 = hashlib.sha256() if not self.config.consider_parents_for_recipe_checksum \
```
ylatuya: I think an If else would be more clear then a online split in several lines

cerbero/config.py
```diff
@@ -449,6 +450,7 @@ def load_defaults(self):
         self.set_property('extra_bootstrap_packages', {})
         self.set_property('binaries_local', None)
         self.set_property('binaries_remote', None)
+        self.set_property('consider_parents_for_recipe_checksum', False)
```
ylatuya: I would name the property by what it does and not how it does it. My  suggestion is strict_cache_invalidation (there can be potentially more things to invalidate the cache in a strict way)

cerbero/utils/__init__.py
```diff
@@ -577,3 +580,16 @@ def is_text_file(filename):
     global TEXTCHARS
     with open(filename, 'rb') as f:
         return len(f.read(1024).translate(None, TEXTCHARS)) <= 10
+
+@lru_cache(maxsize=None)
+def get_class_checksum(clazz):
+    '''
+    Return the SHA256 hash from the source lines of a class.
+    This method uses an LRU cache to avoid calculating the same
```
ylatuya: typo I think, but I am not sure: an LRU -> a LRU

cerbero/config.py
pamarcos: Actually, discussing the name with other team members, they proposed to use something more explicit like this rather than the `strict` name which wasn't clear for some people.

cerbero/utils/__init__.py
pamarcos: Well, this is an internal debate I always have. Since you pronounce LRU as /EL AR IU/, shouldn't it be *an* "EL", since the final N is to help break 2 consecutive vowels?

ylatuya: Me too I never know what is correct

neodesys: an LRU => it's always based on sound, if it sounds like a vowel it's 'an' else it's 'a'.
Curiously in english Y is not a vowel, so if it sounds like 'i' it's 'a'

PR #263, opened 2020-03-24
Lon 1656 2

ylatuya:


recipes/proxy-libintl/0001-Rename-symbols-for-compatibility-with-gettext.patch
```diff
@@ -0,0 +1,101 @@
+From ed92bd68ebbc660e44545be106bfc9897ec75798 Mon Sep 17 00:00:00 2001
```
dfernandezramos: I understand that with this changes I have to revert my changes in windows that rename gettext symbols by adding the "g_libintl_*" thing and let it as it was originally 👍

dfernandezramos: We need to remember to tell @diegopinedo to rebuild again all the dependencies that use these symbols [proxy-libintl, glib, glib-networking, libsoup, gdk-pixbuf] I don't know if there is another recipe that needs a buildone, but I don't think so.

ylatuya: I took care of reverting your gettext changes too in the other PR

PR #266, opened 2020-03-26
OPE-656

neodesys:


cerbero/packages/osx/packager.py
```diff
@@ -302,59 +311,70 @@ def _create_product(self, package_type):
...
                 self.empty_packages[PackageType.RUNTIME].append(p)
+
             if paths[1] is not None:
                 self.packages_paths[PackageType.DEVEL][p] = paths[1]
             else:
                 self.empty_packages[PackageType.DEVEL].append(p)
 
-    def _create_packages_dmg(self):
-        paths = list(self.packages_paths[PackageType.RUNTIME].values())
+            extra_paths += paths[2:]
```
ylatuya: Can't you use the same dictionary self.packages_paths used to accumulate paths ?

cerbero/packages/osx/packager.py
```diff
@@ -302,59 +311,70 @@ def _create_product(self, package_type):
...
 
         m.action(_("Creating image %s ") % dmg_file)
         # create a temporary directory to store packages
-        workdir = os.path.join(self.tmp, "hdidir")
+        workdir = os.path.join(self.tmp, self.package.name)
         os.makedirs(workdir)
         try:
             for p in paths:
-                shutil.copy(p, workdir)
+                shutil.move(p, workdir)
```
ylatuya: This will delete the packages from the output dir.
After a package like the fluendo-sdk-installer is packaged you expect to have the following packages:
fluendo-sdk-installer-1.1.0.740-x86_64-packages.dmg 
fluendo-sdk-installer-1.1.0.740-x86_64.pkg
fluendo-sdk-installer-devel-1.1.0.740-x86_64.pkg

cerbero/packages/osx/packager.py
```diff
@@ -302,59 +311,70 @@ def _create_product(self, package_type):
...
+            extra_paths += paths[2:]
+        return extra_paths
+
+    def _create_packages_dmg(self, paths):
         dmg_file = os.path.join(self.output_dir, self._package_name('-packages.dmg'))
 
         m.action(_("Creating image %s ") % dmg_file)
         # create a temporary directory to store packages
-        workdir = os.path.join(self.tmp, "hdidir")
+        workdir = os.path.join(self.tmp, self.package.name)
```
ylatuya: What's wrong with the previous name? it's created in the temporary directory and indicates it's where the hdi image is going to be created, it's more useful than the package name

cerbero/packages/osx/packager.py
```diff
@@ -124,6 +124,8 @@ def pack(self, output_dir, devel=True, force=False, keep_temp=False, split=True,
 
         self.install_dir = install_dir or self.package.get_install_dir()
         self.version = version or self.package.version
+        if not sdk_version and hasattr(self.package, 'sdk_version'):
```
ylatuya: sdk_version is passed as a parameter because OSXPackager works with a cerbero.packages.package.Package and this class does not have the sdk_version field, only MetaPackage does (https://github.com/fluendo/cerbero/blob/aef435a8b1d5c65b376d445ff846efd97eb8b7f7/cerbero/packages/package.py#L428). So this is checking a condition that would never happen

cerbero/packages/osx/packager.py
```diff
@@ -234,26 +236,32 @@ def pack(self, output_dir, devel=False, force=False, keep_temp=False, split=True
         PackagerBase.pack(self, output_dir, devel, force, keep_temp, split)
 
         self._prepare_pack()
-
-        if isinstance(self.package, MetaPackage):
+        paths = []
+        try:
+            # Create osx-bundle package (self.package is ALWAYS a MetaPackage)
```
ylatuya: The documentation of the class already says this:
"Creates an osx package from a L{cerbero.package.package.MetaPackage}"

cerbero/packages/package.py
```diff
@@ -435,6 +432,13 @@ class MetaPackage(PackageBase):
     @type resources_distribution: string
     @cvar user_resources: folders included in the .dmg for iOS packages
     @type user_resources: list
+    @cvar osx_framework_library: name and link as a (name, relative_path)
+                                 tuple to the Framework library directory
+    @type osx_framework_library: tuple
+    @cvar osx_build_dmg: set to True (default) to move all product packages
```
ylatuya: This changes the actual behaviour that creates always both the pkg and the dmg.

cerbero/packages/package.py
```diff
@@ -434,6 +432,13 @@ class MetaPackage(PackageBase):
     @type resources_distribution: string
     @cvar user_resources: folders included in the .dmg for iOS packages
     @type user_resources: list
+    @cvar osx_framework_library: name and link as a (name, relative_path)
```
ylatuya: A MetaPackage should not produce a macOS Framework, only the SDKPackage should.
What seems to be wrong and might have confused you is this:
  if isinstance(self.package, MetaPackage):
            packager = self._create_framework_bundle_packager()
            self._create_framework_bundle_layout(packager)
            self._create_framework_bundle_package(packager)

cerbero/packages/osx/packager.py
neodesys: In fact, sdk_version is used by fluendo-sdk (which is a package) and fluendo-sdk-installer (which is a meta-package). The issue comes from the initial conception of fluendo-sdk package. The result is that if you build fluendo-sdk package you will have a package installing itself in 1.1.0.xxx (which is the package version), whereas if you install fluendo-sdk-installer package it will install into 1.1 (which is the sdk_version).

Exactly the same files are installed, so it sounds quite incoherent. This is the meaning of this patch, if the package has a sdk_version field ALTHOUGHT it is NOT a meta-package, I read this field.

From the fluendo-sdk package point of view it seems more coherent and it doesn't have impact on other packages, but if you want I can remove it. Then everybody will need to be aware that UNLIKE all other platforms, fluendo-sdk and fluendo-sdk-installer produce different installation paths on Mac.

cerbero/packages/osx/packager.py
```diff
@@ -160,49 +162,59 @@ def _create_package(self, package_type):
...
+        try:
+            tmp, root, resources = self._create_bundle(files, package_type)
+            self.package.pre_build(root)
+            packagebuild = PackageBuild()
+            packagebuild.create_package(root, self.package.identifier(),
+                                        self.package.version, self.package.shortdesc, output_file,
+                                        self._get_install_dir(), scripts_path=resources)
+            return output_file
+        finally:
+            if tmp is not None:
```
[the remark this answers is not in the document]
neodesys: OK, I'm checking it.

cerbero/packages/osx/packager.py
```diff
@@ -160,49 +162,59 @@ def _create_package(self, package_type):
...
+                shutil.copy(in_path, out_path, follow_symlinks=False)
+            if package_type == PackageType.DEVEL:
+                self._create_framework_headers(self.config.prefix, self.include_dirs, root)
+
+            # Copy scripts to the Resources directory
+            os.makedirs(resources)
+            if os.path.isfile(self.package.resources_preinstall):
+                dst = os.path.join(resources, 'preinstall')
+                shutil.copy(self.package.resources_preinstall, dst)
+                shell.new_call(['chmod', '755', dst])
```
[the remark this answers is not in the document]
neodesys: yes better, I will modify it too at other places

cerbero/packages/osx/packager.py
```diff
@@ -160,49 +162,59 @@ def _create_package(self, package_type):
...
+            if os.path.isfile(self.package.resources_postinstall):
+                dst = os.path.join(resources, 'postinstall')
+                shutil.copy(self.package.resources_postinstall, dst)
+                shell.new_call(['chmod', '755', dst])
+            return tmp, root, resources
+        except Exception:
+            shutil.rmtree(tmp)
```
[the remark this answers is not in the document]
neodesys: Ok

cerbero/packages/osx/packager.py
neodesys: yes but the code before was checking is isinstance(self.package, MetaPackage), so I think it's not a bad idea to insist here ;)

cerbero/packages/osx/packager.py
neodesys: no because this dictionary is used at different places expecting only one path and nothing more.
This mechanism expects that all packages only produce 2 outputs: runtime and devel.
In fluendo we may produce more than 2 outputs (like the nuget packages in fluendo-sdk) and I found the idea interesting to be able to archive those extra packages too in the dmg archive in order to distribute them with the sdk installer package itself.

cerbero/packages/osx/packager.py
neodesys: It's a question of first user experience: when you receive your product in a dmg file and you double-click on it, do you prefer having a drive called "hdidir" mounted on your desktop or a drive called "fluendo-sdk-installer" (or whatever is your product name) ?

cerbero/packages/osx/packager.py
neodesys: the dmg archive is just a big zip of the packages. So if you take the dmg archiving option (I put it optional) all those packages are within the dmg archive, it is cumbersome to have them at 2 different places.

Indeed, what the caller expects is what it gets in the post_package() paths array.

Moreover keeping both can create errors during packages creation, one concrete example:
 package signing is done at post_package() stage (we should change this because it should be done during package creation, there are some options on pkgbuild for that, but it's another story). If you create a dmg archive and you let package in place, then the installer package inside the dmg archive will NOT be signed whereas the package outside will be.

That's why to prevent confusion I've simplified the use cases:
- or you take the dmg archive option and this is what you got with all your packages inside the dmg (then you cannot post-process individual packages)
- or you don't take the dmg archive and then you have all your packages in the folder ready for post-processing (fluendo-sdk-installer doesn't use the dmg option for example)

cerbero/packages/package.py
neodesys: With fluendo-sdk packages this behavior is dangerous because it can lead to different packages inside and outside the dmg archive, so it's a good idea to have this behavior switchable.

By default it is at True, the default/actual behavior is safe.

cerbero/packages/package.py
neodesys: This flag was set in Package AND SDKPackage before.
There are lot of incongruities in all this code.
According to where it is used, this is the place it should be.
If there are places where it is used and it should not be, well.... when we've got time it would be a good idea to refactor all this stuff, but for now we cannot fix the whole cerbero in one PR ;)

cerbero/packages/osx/packager.py
ylatuya: Got it, it makes sense :)

PR #267, opened 2020-03-26
COM-1152 Fix recipes reset when not needed

pamarcos:
Commit 382273e1f655c80df6263bb3899324659ded2eab introduced a
regression because `installed_files` is kept to be able to clean
old files when installing the new version of a recipe.
However, this made the recipes to be kept in the status, which
caused too many resets when unnecessary.

### DIFF
--- a/cerbero/build/cookbook.py
+++ b/cerbero/build/cookbook.py
@@ -344,7 +344,7 @@ class CookBook (object):
         @rtype: list
         '''
         recipe = self.get_recipe(recipe_name)
-        return [r for r in list(self.recipes.values()) if recipe.name in r.deps]
+        return list(set([r for r in list(self.recipes.values()) if recipe.name in r.deps]))
 
     def save(self):
         try:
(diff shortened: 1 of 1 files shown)


cerbero/build/cookbook.py
```diff
@@ -411,12 +411,26 @@ def _recipe_status(self, recipe_name):
...
         return self.status[recipe_name]
 
+    def _recipe_is_reset(self, recipe_name):
+        '''
+        We need to use this method to check whether a recipe is reset
+        or not because the installed_files is always kept for previously
+        installed recipes to make sure we purge old files (if needed)
+        when installing again the recipe
+        '''
+        st = self.status[recipe_name]
```
ylatuya: a recipe with that name might not be in the status

cerbero/build/cookbook.py
```diff
@@ -426,7 +440,7 @@ def _reset_recipe_status(self, recipe, change_name, previous, now):
     def _reset_recipe_if_needed(self, recipe):
         # Check that the recipe still exists in cache, because
         # a prior recipe might have reset its status
-        if not recipe.name in self.status:
+        if not recipe.name in self.status or self._recipe_is_reset(recipe.name):
```
ylatuya: checking if recipe.name is in self.status should be part of the is_reset function

cerbero/build/cookbook.py
```diff
@@ -411,12 +411,26 @@ def _recipe_status(self, recipe_name):
                     file_hash=recipe.get_checksum(), installed_files=[])
         return self.status[recipe_name]
 
+    def _recipe_is_reset(self, recipe_name):
+        '''
+        We need to use this method to check whether a recipe is reset
```
ylatuya: This looks more like a comment than documentation for the method itself.
The documentation should explain what the method does, why you use might change over time :)

cerbero/build/cookbook.py
```diff
@@ -467,7 +481,8 @@ def _load_recipes(self):
             # allow safe relocation of the recipes.
             if recipe.__file__ != st.filepath:
                 st.filepath = recipe.__file__
-            recipe.run_func_depending_on_built_version(async_tasks, self._reset_recipe_if_needed, recipe)
+            if not self._recipe_is_reset(recipe.name):
+                recipe.run_func_depending_on_built_version(async_tasks, self._reset_recipe_if_needed, recipe)
```
ylatuya: The method it's calling already checks if self._recipe_is_reset(recipe.name), it's a bit confusing

pamarcos: Actually, it looks weird but it is indeed needed here. If we don't use it, in case the `built_version` is an async task (such as in Git sources), a Future task will be scheduled in order to collect it. That will eventually call `git ls-remote` when it's unnecessary, since we already know the recipe is reset and we don't need to reset it because there was a change in the built_version

ylatuya: That is strange, why would it call it? The function it's deferring is going to check if the recipe is reset  at the very beginning and exit

pamarcos: Previously to calling the method that resets the recipe, you need to already have the proper `built_version` for a given recipe. In case the method to get that is an async method (which in Git it is), you can't do that sequentially. You need to schedule all the tasks and then gather the results. After that, you can call the reset method.

The async built_version command was done to make sure both fridge and the recipes use the proper commit for the names, resolving them by running `git ls-remote` when the repo has not been cloned.

PR #270, opened 2020-04-01
download: Add config to set local servers where cache tarballs

rgonzalezfluendo:


ylatuya:
This serves the same purpose as the mirrors features, why can it be used directly?

ylatuya:
When we discussed this, the approach was to use the mirrors but making it prepend the mirror to the original URL. The community will accept that change as it was proposed as an improvement in my upstream MR but was not finally done that way

pamarcos:
> When we discussed this, the approach was to use the mirrors but making it prepend the mirror to the original URL. The community will accept that change as it was proposed as an improvement in my upstream MR but was not finally done that way

Agree, I commented the same thing to Rubén. Both seem to have a very similar purpose except for the priority. If it's up to me, we could consolidate mirrors by prepending that to the original URL and good to go

ylatuya:
https://github.com/fluendo/cerbero/blob/89e4508717d3c0e53c71c446988483b87c6b59ec/cerbero/utils/shell.py#L459

ylatuya:
See for more context https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/148#note_143810

rgonzalezfluendo:
Ok. I am going to:

1.- revert my change and prepend the mirror to the list of url.
2.- create a cerbero command to automatically syncronize downloaded tarballs.

rgonzalezfluendo:
Usin only the filename in  the mirror is an error:

```
-----> Checksum failed, tarball /opt/fluendo-codecs-0.10-linux-x86_64/sources/local/openh264-1.7.0/v1.7.0.tar.gz moved to /opt/fluendo-codecs-0.10-linux-x86_64/sources/local/openh264-1.7.0/v1.7.0.tar.gz.failed-checksum
Downloading https://github.com/cisco/openh264/archive/v1.7.0.tar.gz

Recipe 'openh264' failed at the build step 'fetch'
Fatal Error: Checksum for /opt/fluendo-codecs-0.10-linux-x86_64/sources/local/openh264-1.7.0/v1.7.0.tar.gz is '1fec931eb5c94279ad219a5b6e0202358e94a93a90cfb1603578c326abfc1238' instead of '9c07c38d7de00046c9c52b12c76a2af7648b70d05bd5460c8b67f6895738653f'

```

`libvpx.recipe` and `openh264.recipe` has the same filename `v1.7.0.tar.gz`.

A solution should be use the complete path in the mirror.

rgonzalezfluendo:
A lot of network errors are happening today. I would like to finish this.

As stated in the previous comment, the basename cannot be used because it matches different recipes. (`libvpx.recipe` and `openh264.recipe` has the same filename `v1.7.0.tar.gz`.)

But if I stop using basename and use the complete path, the official mirror stops working.

Like using two lists with different priorities is not an option, I will try to update the official mirror.

PR #273, opened 2020-04-02
OPE-875 Remove installed_files before installing the new ones

pamarcos:
Previous logic didn't work properly when the files were never actually copied because they were the same.

New step is added: `clean_old` which doesn't run the uninstall step which may be quite slow on some machines (e.g. Windows), but ensures the previous list of installed files is manually removed prior to the install step.

### DIFF
--- a/cerbero/build/oven.py
+++ b/cerbero/build/oven.py
@@ -219,7 +219,7 @@ class Oven (object):
         # list of installed files for this recipe
         # WARNING: the method to automatically detect files will only work
         # when installing recipes not concurrently
-        if BuildSteps.INSTALL in recipe.steps or BuildSteps.POST_INSTALL in recipe.steps:
+        if set([BuildSteps.INSTALL, BuildSteps.POST_INSTALL]).intersection(set(recipe.steps)):
             self._update_installed_files(recipe, tmp)
 
         if recipe.library_type == LibraryType.STATIC:
--- a/cerbero/build/recipe.py
+++ b/cerbero/build/recipe.py
@@ -127,6 +127,7 @@ class BuildSteps(object):
 
     FETCH = (N_('Fetch'), 'fetch')
     EXTRACT = (N_('Extract'), 'extract')
+    CLEAN_INSTALLED_FILES = (N_('Clean installed files'), 'clean_installed_files')
     CONFIGURE = (N_('Configure'), 'configure')
     COMPILE = (N_('Compile'), 'compile')
     INSTALL = (N_('Install'), 'install')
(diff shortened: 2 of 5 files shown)


cerbero/build/oven.py
```diff
@@ -191,6 +191,8 @@ def _cook_recipe(self, recipe, count, total, fridge=None, use_binaries=False, up
                 stepfunc = getattr(recipe, step)
                 if not stepfunc:
                     raise FatalError(_('Step %s not found') % step)
+                if step == BuildSteps.INSTALL[1]:
```
ylatuya, line 194: The oven is dumb, it should only run step functions. We need to away to do this cleanly, either we a new step or adding it in the base class of the install step

pamarcos:
Please @ylatuya take another look to know whether there's something else to do here or not. The fact that this is not merged or fixed yet is blocking COM-1124

cerbero/build/recipe.py
```diff
@@ -127,6 +127,7 @@ class BuildSteps(object):
 
     FETCH = (N_('Fetch'), 'fetch')
     EXTRACT = (N_('Extract'), 'extract')
+    CLEAN_OLD = (N_('Clean old'), 'clean_old')
```
neodesys, line 130: I would have call this clean_old_files or clean_obsolete_files because clean_old only is not super explicit ;)

cerbero/commands/uninstall.py
```diff
@@ -69,20 +69,14 @@ def run(self, config, args):
...
-
+            cookbook.recipe_remove_installed_files(recipe_name)
             cookbook.clean_recipe_status(recipe_name)
 
             # Lastly, remove the build directory
             build_dir = os.path.join(config.sources, recipe.package_name)
             if os.path.isdir(build_dir):
                 shutil.rmtree(build_dir)
 
-        shell.remove_empty_dirs(config.install_dir)
```
neodesys, line 85: why did you remove this one ? it's prettier to remove empty dirs, or maybe there are side effects ?

cerbero/build/recipe.py
pamarcos: That makes sense and it's easily tackled. I'll rename it to `clean_old_files` then

cerbero/commands/uninstall.py
pamarcos: Yep, indeed it has some nasty side effects. For instance, it may remove the `aclocal` dir that cerbero creates (even though it's emtpy). This causes the `autoools` recipe to not be able to install anything within that dir because it doesn't exist. One would expect an installation process to create the dirs that it needs, but that doesn't seem to be the case.

So, in the end I resigned and decided not to remove empty directories because of these side effects. We'll end up having empty and unnecessary dirs, but the alternative is recipes not installing correctly, so...

neodesys: Effectively it is better to no touch them ;)

cerbero/commands/uninstall.py
```diff
@@ -69,20 +69,14 @@ def run(self, config, args):
             # As a fallback, in case the uninstall target does not exist,
             # or in case it misses some of the installed files, remove
             # them manually
-            installed_files = cookbook.recipe_installed_files(recipe_name)
-            for f in [os.path.join(config.install_dir, f) for f in installed_files]:
-                # os.path.exists returns False for broken symbolic links
-                if os.path.exists(f) or os.path.islink(f):
-                    os.remove(f)
-
+            cookbook.recipe_remove_installed_files(recipe_name)
```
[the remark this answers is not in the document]
pamarcos: Good point. PR to address it: https://github.com/fluendo/cerbero/pull/274/

PR #305, opened 2020-09-10
build-tools: Install meson as flat egg

cfoch:
In ubuntu xenial, meson is not recognized inside
flu-codec-sdk shell environment

Issue: OCP-1293

### DIFF
--- a/recipes/build-tools/meson.recipe
+++ b/recipes/build-tools/meson.recipe
@@ -23,8 +23,14 @@ class Recipe(recipe.Recipe):
             prefix = str(PurePath(self.config.prefix))
         else:
             prefix = self.config.prefix
-        shell.new_call([self.config.python_exe, 'setup.py', 'install', '--prefix', prefix],
-                       cmd_dir=self.build_dir, env=self.env, logfile=self.logfile)
+
+        cmd = [self.config.python_exe, 'setup.py', 'install',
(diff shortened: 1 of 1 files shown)


cfoch:
It works as for what I have tested, but I am not sure at all if these options are safe.

pamarcos:
LGTM. According to https://pypi.org/project/setuptools/1.0/:

>Made --single-version-externally-managed automatic when --root is used, so that most system packagers won’t require special support for setuptools.

So we could use just the `--root` parameter, which along with `--prefix` sets everything in the correct dir: `/opt/fluendo-codecs-0.10-linux-x86/build-tools/lib/python3.8/site-packages` as opposed to `/opt/fluendo-codecs-0.10-linux-x86/build-tools/lib/python3.8/site-packages/meson-0.52.0-py3.8.egg/mesonbuild`
CC @ylatuya what do you think?

cfoch:
I added a change to fix an issue in macOS.

pamarcos:
I'm closing this PR since the fix is already included in the work related to OCP-1641. I wanted to talk to you before merging since you are the one who did the fix, but we wanted to merge this ASAP to start doing the releases. I gave you credit on the commit itself: https://github.com/fluendo/cerbero/commit/0680c1acea917cbe6456aef458b6074abce4ddcf

PR #330, opened 2021-02-24
Recipe module for explicit module name resolution

JohnJudd:
This allows to infer details about the recipe using introspection
outside of the recipe folder more accurately. For example; before
this change using the function inspect.getmro to get the path of a
recipe in the recipes folder from it's class; making this call from
a function outside of the recipes  folder, it will return the class
refreence custom.GStreamer. The problem is if there is another
custom.py in the scope you are calling getmro from. The function
inspect.getmro will do a best guess and may not reference the
correct custom.py. So, by turning the recipes folder in to a module
we get the full class name resolution of recipes.custom.GStreamer;
instead of custom.GStreamer. This avoids any confusion to which
custom.py we are referring to.

### DIFF
--- a/recipes/gst-editing-services-1.0.recipe
+++ b/recipes/gst-editing-services-1.0.recipe
@@ -1,5 +1,6 @@
 # -*- Mode: Python -*- vi:si:et:sw=4:sts=4:ts=4:syntax=python
 from cerbero.utils import shell
+from recipes import custom
 
 class Recipe(custom.GStreamer):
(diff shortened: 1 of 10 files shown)


pamarcos (approved):
Excellent work, John! :ok_hand: 
Only a fussy comment to clarify the commit message: it's `inspect.getsourcelines` the one that in the end guesses the file path for a class, based on its module, not `inspect.getmro`. The latter is used to collect the list of parents for a given class

PR #335, opened 2021-03-24
source: git: Avoid cloning common sources multiple times

cfoch:
In some cases, multiple recipes in a cbc may point to the same
git repo. The previous behaviour was to fetch for if recipe. In
this commit, when the repo_dir among multiple recipes is the
same, then only one of them will be fetched and the other
ignored.

There are two cases to ignore fetching a recipe:
 - The recipe was already fetch according the recipe status
 - There is other similar recipe fetching

An asyncio condition is used per group of recipes pointing to a
common git source. It is used to prevent the fetch step done if
there is an-in-progress fetch (for its source); so if the
in-progress feth fails, the others waiting are not considered
fetched in the cache file.

Issue: OCP-1914

### DIFF
--- a/cerbero/build/source.py
+++ b/cerbero/build/source.py
@@ -16,6 +16,7 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
+import asyncio
 import os
 import shutil
 import tarfile
@@ -408,13 +409,80 @@ class Git (GitCache):
     Source handler for git repositories
     '''
 
+    # Maps recipes with common sources share to the same fetch_done condition.
+    _fetch_done_cond_map = {}
+
     def __init__(self):
         GitCache.__init__(self)
         if self.commit is None:
             # Used by recipes in recipes/toolchain/
             self.commit = 'origin/sdk-%s' % self.version
+
+        self._fetching = False
         # For forced commits in the config
         self.commit = self.config.recipe_commit(self.name) or self.commit
+
+    def _get_source_id(self):
+        return (self.repo_dir, tuple(self.remotes.items()))
+
+    def _update_fetch_done_cond(self):
+        source_id = self._get_source_id()
+        if source_id in Git._fetch_done_cond_map:
+            return
+        lock = asyncio.Lock()
+        Git._fetch_done_cond_map[source_id] = asyncio.Condition(lock)
+
+    def _get_git_recipes(self):
+        recipes = self.config.cookbook.get_recipes_list()
+        return [r for r in recipes if issubclass(r.stype, SourceType.GIT)]
+
+    def _already_fetched(self, common_source_recipes):
+        for recipe in common_source_recipes:
+            if self.config.cookbook.step_done(recipe.name, 'fetch'):
+                return True
+        return False
+
+    def _other_fetching(self, common_source_recipes):
+        for recipe in common_source_recipes:
+            if recipe == self:
+                continue
+            if recipe._fetching:
+                return True
+        return False
+
+    async def fetch(self, **kwargs):
+        self._fetching = True
+
+        git_recipes = [r for r in self.config.cookbook.get_recipes_list()
+                           if issubclass(r.stype, SourceType.GIT)]
+        common_source_recipes = [r for r in git_recipes
+                                     if self._get_source_id() == r._get_source_id()]
+        already_fetched = self._already_fetched(common_source_recipes)
+        other_fetching = self._other_fetching(common_source_recipes)
+
+        self._update_fetch_done_cond()
+        fetch_done = Git._fetch_done_cond_map[self._get_source_id()]
+
+        if already_fetched:
+            m.log(_("Ignoring recipe %s: was already fetch.") % self, get_logfile(self))
+            self._fetching = False
+            return
+
+        if other_fetching:
+            m.log(_("Ignoring recipe %s: other similar is already fetching.") % self, get_logfile(self))
+
+            async with fetch_done:
+                await fetch_done.wait()
+            self._fetching = False
+            return
+
+        try:
+            await super().fetch(**kwargs)
+        finally:
+            self._fetching = False
+
+        async with fetch_done:
+            fetch_done.notify_all()
 
     async def extract(self):
         if os.path.exists(self.build_dir):
--- a/cerbero/utils/git.py
+++ b/cerbero/utils/git.py
@@ -286,15 +286,10 @@ async def local_checkout(git_dir, local_git_dir, commit, logfile=None):
     @param commit: the commit to checkout
     @type commit: false
     '''
-    # reset to a commit in case it's the first checkout and the masterbranch is
-    # missing
     branch_name = 'cerbero_build'
-    shell.call('%s reset --hard %s' % (GIT, commit), local_git_dir, logfile=logfile)
-    shell.call('%s branch %s' % (GIT, branch_name), local_git_dir, fail=False, logfile=logfile)
-    await shell.async_call('%s checkout %s' % (GIT, branch_name), local_git_dir, logfile=logfile)
-    await shell.async_call('%s reset --hard %s' % (GIT, commit), local_git_dir, logfile=logfile)
-    await shell.async_call('%s clone %s -s -b %s .' % (GIT, local_git_dir, branch_name),
-                           git_dir, logfile=logfile)
+    await shell.async_call([GIT, 'checkout', commit, '-B', branch_name], local_git_dir, logfile=logfile)
+    await shell.async_call([GIT, 'clone', local_git_dir, '-s', '-b', branch_name, '.'],
+               git_dir, logfile=logfile)
     ensure_user_is_set(git_dir, logfile=logfile)
     await submodules_update(git_dir, local_git_dir, logfile=logfile)
 


cerbero/utils/git.py
```diff
@@ -290,7 +290,7 @@ async def local_checkout(git_dir, local_git_dir, commit, logfile=None):
     # missing
     branch_name = 'cerbero_build'
     shell.call('%s reset --hard %s' % (GIT, commit), local_git_dir, logfile=logfile)
-    shell.call('%s branch %s' % (GIT, branch_name), local_git_dir, fail=False, logfile=logfile)
+    shell.call('%s branch %s %s' % (GIT, branch_name, commit), local_git_dir, fail=False, logfile=logfile)
```
pamarcos, line 293: Why did you need to add the commit here? I'm guessing because you need to ensure the commit here, but I wonder if this may cause any regression because the commit itself was never used. Maybe `cerbero_build` was always used for a reason. In any case, commit doesn't seem to be a boolean, so we need to update the docu for the function

cerbero/build/source.py
```diff
@@ -408,14 +411,93 @@ class Git (GitCache):
...
+        lock = asyncio.Lock()
+        Git._fetch_done_cond_map[source_id] = asyncio.Condition(lock)
+
+    def _get_fetch_done_cond(self):
+        source_id = self._get_source_id()
+        return Git._fetch_done_cond_map[source_id]
+
+    def _get_git_cache_recipes(self):
+        recipes = self.config.cookbook.get_recipes_list()
+        return iter(r for r in recipes if r.stype == SourceType.GIT)
```
pamarcos, line 446: `GitExtractedTarball` inherits from `Git`. So, we need to use `issubclass(r.stype, SourceType.GIT)` rather than checking only for `SourceType.GIT`. However, I've seen some checks checking explicitly for `SourceType.GIT or SourceType.GIT_TARBALL`, so maybe we should use that instead. IMHO,I prefer consistency and whatever we do, we modify it so that either everything checks for inheritance, or everything checks for one or the other stype.

pamarcos (changes requested):
Overall, excellent work. The changes achieve the goal.
`./cerbero-uninstalled -t -c projects/fluffgst/patch/fluffgst-patch-release-1.0.cbc fetch`
went from almost 7min to 50secs. The fewer the cores, the fewer the parallel fetches, the higher the gain. That's very good news :+1: 

Now, I think the solution is somehow convoluted and not easy to grasp at first sight. I know what it does is not precisely easy, but I'd like to try push for a simpler implementation. I'm talking particularly about the need of fetch_done, already_fetch and other_fetching functions.

cerbero/build/source.py
cfoch: I think there's some magic metaclass, that it makes the recipe to automagically inherit from the corresponding value (a class) set in stype (which may be SourceType.GIT, SourceType.GIT_TARBALL, ...). So I think it's the same, but not sure :-)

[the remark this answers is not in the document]
cfoch: Ok, let's go with a list. I wanted to prevent iterating twice, but it's cheap. Also, if one is not familiar with generators, it will be error prone.

cfoch: Ah... this is true, but I should use `issubclass(r.stype, SourceType.GIT)`, yes.

cerbero/build/source.py
```diff
@@ -408,14 +411,93 @@ class Git (GitCache):
...
+
+    def _common_source_recipes(self, target_recipe):
```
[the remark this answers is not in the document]
cfoch: uneeded, yes

cerbero/build/source.py
```diff
@@ -408,14 +411,93 @@ class Git (GitCache):
...
+        recipes = self.config.cookbook.get_recipes_list()
+        git_recipes = iter(r for r in recipes if r.stype == SourceType.GIT)
+
+        target_source_id = target_recipe._get_source_id()
+        for recipe in self._get_git_cache_recipes():
+            if target_source_id == recipe._get_source_id():
+                yield recipe
+        yield from ()
+
+    def _already_fetch(self):
```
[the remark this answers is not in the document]
cfoch: > Now, I think the solution is somehow convoluted and not easy to grasp at first sight. I know what it does is not precisely easy, but I'd like to try push for a simpler implementation. I'm talking particularly about the need of fetch_done, already_fetch and other_fetching functions.

The simpler implementation that I can think of is to address this from the fetch command, instead of the impl.

cerbero/utils/git.py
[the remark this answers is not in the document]
cfoch: I need to double check if something else may be affected. But this is needed because fetch will end up with HEAD pointing to the commit of the fetched recipe, and in the extract step, other recipes specifying other commits, will clone pointing to the wrong commit (the commit of the fetched source).

cerbero/build/source.py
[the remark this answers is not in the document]
pamarcos: Right. The issue is that the implementation you propose here serves both for fetch (which runs in parallel by default with async) but also for build command which will run fetch if needed. The fetch implementation would be much much simpler because you know beforehands which recipes you are going to fetch. That'd avoid the lock mechanism that you implement here. But of course, on the other side it wouldn't be used for build command.

cerbero/build/source.py
```diff
@@ -408,14 +409,103 @@ class Git (GitCache):
...
+
+    def __getattr__(self, attr):
+        if attr == "prepare":
+            def _prepare():
+                # Set the fetch condition object after the default prepare
+                # method, as some recipes (properly) set/change remotes or
+                # repo_dir attributes from the prepare method, and the
+                # _update_fetch_done_cond method needs this info.
+                self.__git_prepare()
+                self._update_fetch_done_cond()
```
pamarcos, line 440: I think this can be called in the fetch instead of doing this complicated thing just to make sure we do it after the prepare. Please give it a thought to see if we can make it simpler

cerbero/build/source.py
```diff
@@ -408,14 +409,103 @@ class Git (GitCache):
...
+                continue
+            if recipe._fetching:
+                return True
+        return False
+
+    async def fetch(self, **kwargs):
+        self._fetching = True
+
+        fetch_done = self._get_fetch_done_cond()
+        already_fetched = self._already_fetched()
```
pamarcos, line 485: We're calling `_common_source_recipes` in both methods, so I think as an idea we could call it once and then pass it as a parameter to `_already_fetched` and `_other_fetching`. Considering it iterates through all recipes, I think it's worth it.

pamarcos (changes requested):

cerbero/build/source.py
```diff
@@ -408,14 +409,85 @@ class Git (GitCache):
...
+        already_fetched = self._already_fetched(common_source_recipes)
+        other_fetching = self._other_fetching(common_source_recipes)
+
+        self._update_fetch_done_cond()
+        fetch_done = Git._fetch_done_cond_map[self._get_source_id()]
+
+        if already_fetched:
+            m.log(_("Ignoring recipe %s: was already fetch.") % self, get_logfile(self))
+            self._fetching = False
+            print("already fetched")
```
pamarcos, line 470: seems like a debug trace

cerbero/build/source.py
```diff
@@ -408,14 +409,85 @@ class Git (GitCache):
...
+        if already_fetched:
+            m.log(_("Ignoring recipe %s: was already fetch.") % self, get_logfile(self))
+            self._fetching = False
+            print("already fetched")    
+            return
+
+        if other_fetching:
+            m.log(_("Ignoring recipe %s: other similar is already fetching.") % self, get_logfile(self))
+
+            print("waiting")
```
pamarcos, line 476: seems like a debug trace

cerbero/build/source.py
```diff
@@ -408,14 +409,85 @@ class Git (GitCache):
...
+            async with fetch_done:
+                await fetch_done.wait()
+            self._fetching = False
+            return
+
+        print("fetching")
```
pamarcos, line 482: seems like a debug trace

pamarcos (changes requested):

cerbero/utils/git.py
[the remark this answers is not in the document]
pamarcos: @cfoch I've taken a look at this and I've found two things:
1. We don't need to pass the commit when creating the branch because the previous `reset --hard` already makes sure we're pointing to the right commit
2. We have duplicated calls (from when I move to async). We can consolidate them and do the same thing the community cerbero is doing: https://gitlab.freedesktop.org/gstreamer/cerbero/-/blob/master/cerbero/utils/git.py#L223

```
    await shell.async_call([GIT, 'checkout', commit, '-B', branch_name], local_git_dir, logfile=logfile)
    await shell.async_call([GIT, 'clone', local_git_dir, '-s', '-b', branch_name, '.'],
               git_dir, logfile=logfile)
```

Please change this, rebase with autosquash and we're good to merge. Afterwards, you'll need to update https://github.com/fluendo/fluendo-cerbero/pull/1029 to the new cerbero commit (the merge commit), and then merge the fluendo-cerbero one.

pamarcos (changes requested):

cerbero/build/source.py
```diff
@@ -408,14 +409,82 @@ class Git (GitCache):
...
+        Git._fetch_done_cond_map[source_id] = asyncio.Condition(lock)
+
+    def _get_git_recipes(self):
+        recipes = self.config.cookbook.get_recipes_list()
+        return [r for r in recipes if issubclass(r.stype, SourceType.GIT)]
+
+    def _already_fetched(self, common_source_recipes):
+        for recipe in common_source_recipes:
+            status = self.config.cookbook.status.get(recipe.name)
+            if status and 'fetch' in status.steps:
```
pamarcos, line 442: You can actually use ` self.cookbook.step_done(recipe.name, 'fetch')` instead

pamarcos (approved):

PR #344, opened 2021-05-26
OCP-2206: git: Prevent fetching submodules if no git modules present

cfoch:
It skips an error when fetching submodules (which is the default behaviour) in repos not using submodules.
Supporting fail, would involve touching too much parts of fluendo-cerbero, and we prefer to avoid that and keep it simple.

### DIFF
--- a/cerbero/utils/git.py
+++ b/cerbero/utils/git.py
@@ -19,9 +19,10 @@
 import os
 import time
 import shutil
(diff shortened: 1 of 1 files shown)


pamarcos (approved):
Nice. Remember to squash commits before merging. After that, you'll have to update fluendo-cerbero sumodule as usual :+1:

PR #358, opened 2021-11-08
gobject-introspection: Fix compilation in Arch Linux

carlosfg-flu:
Arch uses python3 by default, so there is no need to modify the
environment. Furthermore, there is a bug here because the attribute
"self.config_sh" is not defined when used.


rgonzalezfluendo (approved):

carlosfg-flu:
I just realize that I didn't use my Fluendo mail here. If is OK for you @rubenrua, I'll fix it.

PR #359, opened 2022-01-07
OSX: don't try to relocate ar archives

aslobodeniuk:
It fails with message "*** is not a Mach-O file"

### DIFF
--- a/cerbero/tools/osxrelocator.py
+++ b/cerbero/tools/osxrelocator.py
@@ -118,8 +118,17 @@ class OSXRelocator(object):
 
     @staticmethod
     def is_mach_o_file(filename):
-        return '.dylib' in os.path.splitext(filename)[1] or \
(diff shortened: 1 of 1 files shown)


rgonzalezfluendo:
Where do you get the issue? Can you add the link to the CI?

aslobodeniuk:
> Where do you get the issue? 

Dolby SDK has a static library (ar archive) that I install to the prefix (together with IPPs that also have ar archives) with a recipe.
Then when the plugin is built, it fails on the step relocate_osx_libraries, 

Its strange, but I've just watched all the CI builds of my branch and havent found that failure. Probably because at some moments I did git commit --amend..

But I've started another build that does not include changes from this PR.

http://ci.fluendo.com:8080/blue/organizations/jenkins/flu-plugins/detail/OCP-3333-osx-fail-demo/1/pipeline

cfoch (approved):

PR #364, opened 2022-04-27
vala-m4: Use "gitlab.gnome.org" as source instead of "download.gnome.org"

carlosfg-flu:
The URL https://download.gnome.org/sources/vala/0.35/vala-0.35.2.tar.xz is
not working, fetch directly from the git repository instead.

### DIFF
--- a/recipes/build-tools/vala-m4.recipe
+++ b/recipes/build-tools/vala-m4.recipe
@@ -9,10 +9,9 @@ class Recipe(recipe.Recipe):
     """
     name = 'vala-m4'
     version = '0.35.2'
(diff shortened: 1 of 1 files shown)


carlosfg-flu:
Maybe is wise to wait a couple of days, just in case https://download.gnome.org/sources/vala/0.35/vala-0.35.2.tar.xz start to work again.

carlosfg-flu:
> Maybe is wise to wait a couple of days, just in case https://download.gnome.org/sources/vala/0.35/vala-0.35.2.tar.xz start to work again.

It's working now :facepalm:

PR #369, opened 2022-09-08
OCP-3457 distros: Fix CentOS 7 allowance

cfoch:
Before the pattern was:
  ('CentOS Linux', '7.9.2009', 'Core')

But after image update it is now:
  ('CentOS Linux', '7', 'Core')


rgonzalezfluendo (approved):

rgonzalezfluendo:
my two cents:
```
-But after image update it is now:
+But after installing Python 3.8 from centos-release-scl it is now
```

aslobodeniuk (approved):

PR #375, opened 2022-11-07
build: Makefile: Do not clear CONFIG_SITE unless distro is OPEN_SUSSE

carlosfg-flu:
CONFIG_SITE could be useful in other cases, do not restring its use.


carlosfg-flu:
We go with another option: https://github.com/fluendo/cerbero/pull/374

PR #396, opened 2024-10-30
cmake: Configure using

cfoch:
Ref #RDI-2128


carlosfg-flu (approved):
LGTM

rgonzalezfluendo (changes requested):
This PR should follow this [use case](https://github.com/fluendo/git-upstream-workflow/blob/main/docs/workflows.md#add-a-new-commit-into-a-pending-feature-because-of-an-internal-requirement-without-conflict) using the [emscripten branch](https://github.com/fluendo/cerbero/tree/emscripten)

PR #404, opened 2024-11-26
Emscripten fix bootstrapper base and x265

rgonzalezfluendo:


cfoch:
I am sure this https://github.com/fluendo/cerbero/pull/405 compiled. I am not sure this compiles, too.

cfoch:
I will approve, but I do not understand why we add "fixup" commits.

PR #407, opened 2025-01-17
OCP-6055 update 'imp' module to 'importlib'

pgsancho:
For Python 3.12+, 'imp' module is called 'importlib'. flu-plugins' CI was updated to 3.12, so it was breaking it.

Issue: OCP-6055

### DIFF
--- a/cerbero/bootstrap/site-patch.py
+++ b/cerbero/bootstrap/site-patch.py
@@ -25,7 +25,7 @@ def __boot():
                 break
         else:
             try:
-                import imp  # Avoid import loop in Python 3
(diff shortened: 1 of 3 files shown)


cerbero/bootstrap/site-patch.py
```diff
@@ -25,7 +25,7 @@ def __boot():
                 break
         else:
             try:
-                import imp  # Avoid import loop in Python 3
+                import importlib as imp  # Avoid import loop in Python 3
```
rgonzalezfluendo, line 28: This is not the right solution:

`AttributeError: module 'importlib' has no attribute 'find_module'`

PR #409, opened 2025-02-04
OCP-6113

carlosfg-flu:
Do not install subprojects


rgonzalezfluendo:
Done upstream in https://gitlab.freedesktop.org/gstreamer/cerbero/-/commit/5278bc515

rgonzalezfluendo (approved):

PR #417, opened 2025-05-12
Add 'hbbtv' variant

aslobodeniuk:
This variant enables maximum compatibility with old versions of the browsers and maximum binary size reduction

Issue: RDI_2506


cfoch (approved):
LGTM at first glance

PR #418, opened 2025-05-14
gst-plugins-rs: support building with old binutils versions

diegonieto:
handle the export parameter in ld based on the distribution


rgonzalezfluendo:
Please use guw.

Also create this MR upstream to get early feedback.

diegonieto:
> Also create this MR upstream to get early feedback.

MR created at https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/1812. Closing this.

PR #420, opened 2025-05-28
web: don't disable dash plugin

aslobodeniuk:
Issue: CS_3298

### DIFF
--- a/recipes/gst-plugins-bad-1.0.recipe
+++ b/recipes/gst-plugins-bad-1.0.recipe
@@ -616,7 +616,6 @@ class Recipe(custom.GStreamer):
             self.disable_plugin('aes', 'effects')
             self.disable_plugin('assrender', 'codecs', dep='libass')
             self.disable_plugin('closedcaption', 'effects', dep='pango')
-            self.disable_plugin('dash', 'codecs', dep='libxml2')
             self.disable_plugin('dtsdec', 'codecs_restricted', option='dts', dep='libdca')
             self.disable_plugin('dtls', 'net')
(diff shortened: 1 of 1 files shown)


recipes/gst-plugins-bad-1.0.recipe
```diff
@@ -630,7 +629,7 @@ class Recipe(custom.GStreamer):
             self.meson_options['sctp-internal-usrsctp'] = 'disabled'
             self.disable_plugin('soundtouch', 'effects', dep='soundtouch')
             self.disable_plugin('spandsp', 'codecs', dep='spandsp')
-            self.disable_plugin('smoothstreaming', 'codecs', dep='libxml2')
+            self.disable_plugin('smoothstreaming', 'codecs')
```
cfoch, line 632: We should have some way to enable or disable plugins without editing cerbero.

aslobodeniuk: I agree, but that's out of scope: I didn't invent this recipe, it was already disabling some plugins for web:
```
        if self.config.target_platform == Platform.WEB:
            self.disable_plugin('aes', 'effects')
            self.disable_plugin('assrender', 'codecs', dep='libass')
```
https://github.com/fluendo/cerbero/commit/4ce83e6c10fbfeeb4809eef09528fe66f5b2359b

Here I'm basically just fixing up this commit from the `emscripten` feature (the PR is to).

Also "We should have some way to enable or disable plugins without editing cerbero" sounds more a general request for cerbero not even gst.wasm

